PR #5, opened 2018-05-13
UI base url support including `kubectl proxy` special case

mthx:
More invasive than would be ideal, as react router requires paths to be absolute.

The `kubectl proxy` support is based on the `x-forwarded-uri` header it sets.


mthx:
This obviously complicates things a bit, but I think it's worth considering having argo-ui work via kubectl proxy by default. What do people think?

mthx:
I'd like to play with this a bit more before considering merging (I need to test the port-forward case), but feedback very welcome.

mthx:
Happy now.  I thought I'd introduced https://github.com/argoproj/argo/issues/858 but that looks to be a separate issue.

src/api/app.ts
```diff
@@ -112,16 +112,30 @@ export function create(
         streamServerEvents(req, res, logsSource, (item) => item.toString());
     });
 
+    const uiBaseHref = (req: express.Request) => {
```
alexmt: I think `kubectl proxy` itself update href value for index.html. I've tried to revert changes in app.ts and argo ui still works with `kubectl proxy`. Please revert changes in app.ts (unless I'm missing something)

src/app/workflows/components/workflows-list/workflows-list.tsx
```diff
@@ -49,16 +50,16 @@ class Component extends React.Component<Props, any> {
             selectedValues: this.props.phases,
             selectionChanged: (phases) => {
                 const query = phases.length > 0 ? '?' + phases.map((phase) => `phase=${phase}`).join('&') : '';
-                this.appContext.router.history.push(`/workflows${query}`);
+                this.appContext.router.history.push(uiUrl(`workflows${query}`));
             },
         };
         return (
-            <Page title='Workflows' toolbar={{filter, breadcrumbs: [{ title: 'Workflows', path: '/workflows' }]}}>
+            <Page title='Workflows' toolbar={{filter, breadcrumbs: [{ title: 'Workflows', path: uiUrl('workflows') }]}}>
```
alexmt: Please apply same change here https://github.com/argoproj/argo-ui/blob/HEAD/src/app/workflows/components/workflow-details/workflow-details.tsx#L66

src/api/app.ts
mthx: Definitely me missing something :-)

I've reverted changes in app.tsx except for the `{index: false}` option to `express.static` which is necessary to use the modified index HTML when `/` is accessed directly (a6a03b20636e0edc0aa5322c0d065a59ae40d4ae).

PR #14, opened 2018-09-06
force namespace isolation

DennisRutjes:
Argo is capable of operating in a single namespace. 

If you deploy argo-ui the same way (without clusterrole/clusterrolebindings just normal role/rolebindings) workflows cannot be accessed.

I created an extra environment configuration option "FORCE_NAMESPACE_ISOLATION" to allow this to function on the UI for a singlenamespace.

Hopefully this will help others to use argo-ui in a single namespace

With kind regards,

Dennis Rutjes

### DIFF
--- a/src/api/main.ts
+++ b/src/api/main.ts
@@ -11,6 +11,7 @@ app.create(
   argv.uiBaseHref || '/',
   argv.inCluster === 'true',
   argv.namespace || 'default',
+  argv.forceNamespaceIsolation || 'false',
   argv.instanceId || undefined,
   argv.crdVersion || 'v1alpha1',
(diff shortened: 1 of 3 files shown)


src/api/main.ts
```diff
@@ -11,6 +11,7 @@ app.create(
   argv.uiBaseHref || '/',
   argv.inCluster === 'true',
   argv.namespace || 'default',
+  argv.forceNamespaceIsolation || 'false',
```
alexmt: Please change it to `argv.forceNamespaceIsolation === 'true'` . Otherwise UI always run in force namespace isolation mode.

PR #16, opened 2018-09-28
Add more logging

iterion:
Adding express-winston turned out to be a pretty quick and easy way to add a bit more logging to the UI. I left the current error logging as is, but it might make sense to just remove it and use the error reporting that express-winston provides.


alexmt:
Thank you @iterion !

PR #17, opened 2018-10-10
Fix pointer usage: from top-left to middle

FurtadoT:
Hello!

While changing node size, I noticed that the values {x, y} returned by dagre are the center point of node/edge, as mentioned in [dagre wiki](https://github.com/dagrejs/dagre/wiki#output-graph):

> node, edge | x | For nodes, the x-coordinate for the center of the node. For edges the x-coordinate for the center of the edge label.
> 
> node, edge | y | For nodes, the y-coordinate for the center of the node. For edges the y-coordinate for the center of the edge label.

The code was using these values as top left point when positioning nodes. Some CSS translations and margins were added to correct those positioning, even though not necessary when top-left coordinates are properly calculated.

The translations were hard-coded to the currently used node width and height, making it impossible to change them without breaking the UI.

So I took the opportunity to create this PR changing the way that positions are calculated.

Hope it helps :)

### DIFF
--- a/src/app/workflows/components/workflow-dag/workflow-dag.scss
+++ b/src/app/workflows/components/workflow-dag/workflow-dag.scss
@@ -4,7 +4,6 @@
 
 .workflow-dag {
     position: relative;
-    overflow: hidden;
 
     &__line {
         position: absolute;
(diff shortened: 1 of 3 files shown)


alexmt:
Thank you for contribution @FurtadoT ! This looks much cleaner

PR #18, opened 2018-10-15
fix: fix get logs api, get namespace from parameters instead of query

jackfengji:
Logs is not showing if workflow is in another namespace in argo-ui from v2.2.1


alexmt:
👍 thank you @jackfengji

PR #19, opened 2018-12-15
Fix double scrollbars in workflow details

ObeA:
This commit should fix double scrollbars showing in the workflow details page. This could happen when you've got a very large workflow.

Before and after this PR:
Example


robintanner:
I understand there are supposed to be additional components to the UI in the npm package, but I haven't been able to figure out how to get this installed on AWS. Is there a best practice for replacing/upgrading the existing UI install from github with the npm version?

alexec:
Do you still need this please, if so - please re-open?

ObeA:
The double scrollbars still seem to be there in the latest version.

PR #23, opened 2019-03-14
Added support to decode the compressed nodes on workflow

sarabala1979:


src/api/app.ts
```diff
@@ -111,7 +112,7 @@ export function create(
             stream.on('close', () => observer.complete());
             stream = stream.pipe(new JSONStream());
             stream.on('data', (data) => data && observer.next(data));
-        });
+        }).flatMap(deCompressNodes);
```
alexmt: The `updatesSource` returns Observable of `{ object: models.Workflow, eventType: string  } `, so deCompressNodes should be executed on `object` field.

I know that `flatMap` works with function which returns Observable, not Promise. Does it work with Promise too?

If you change apply both changes that code might looks like this:

```
}).flatMap((change) => Observable.fromPromise(deCompressNodes(change.object).then((workflow) => ({...change, object: workflow}))));
```

alexmt:
Please use fork for next PR

src/api/app.ts
```diff
@@ -98,11 +100,10 @@ export function create(
     }));
 
     app.get('/api/workflows/:namespace/:name',
-        async (req, res) => serve(res, () => (forceNamespaceIsolation ? crd.ns(namespace) : crd.ns(req.params.namespace)).workflows.get(req.params.name)));
```
alexmt: Sorry, missed in during first review. Same change is required for `app.get('/api/workflows')`

PR #31, opened 2019-07-15
Always draw hidden items of dropdown

dtaniwaki:
The dropdown menu was not rendered inside the window like below when I first clicked the anchor.

Screen Shot 2019-07-15 at 12 03 59 PM

If I clicked the anchor again, it was rendered inside.

I found the closed dropdown menu didn't have child elements and failed to get the size of them to correctly render the menu inside.


alexmt:
Lazy rendering is used on argocd app details page. It is required because for nodes action menu since we need to make an API call to load available node actions. 

The error on screenshot was fixed by https://github.com/argoproj/argo-ui/commit/20bf3cbe3db2c60f5882bd95c153c538e39a0b7b . Fix is available in v1.1 (it is not yet released)

dtaniwaki:
@alexmt I tried the latest Argo CD at 5e4f327 and it looks good.

PR #33, opened 2019-07-25
Removes base width so that components expand to fuller width

alexec:


src/app/shared/styles/elements/containers.scss
```diff
@@ -1,6 +1,5 @@
 .argo-container {
     position: relative;
-    width: $basePageWidth + 160;
```
alexmt: The .argo-container  and fixed-width are meant to limit width. Instead of removing width here we should just stop using .argo-container in places where it is not appropriate.

alexec: ok, let me try that

alexec:
that works - closing

PR #34, opened 2019-08-15
Fix history import

vokracko:
Fix deprecated history import

> Warning: Please use `require("history").createBrowserHistory` instead of `require("history/createBrowserHistory")`. Support for the latter will be removed in the next major release.

and

> app.tsx:10 Uncaught TypeError: createBrowserHistory_1.default is not a function


alexec:
@vokracko do you still need this please, if so - please re-open?

PR #59, opened 2020-04-06
feat: Allow JSX.Element in nav-bar icon

simster7:


simster7:
My original intention was to allow icon stacking by passing in a `<span>` class. See: https://fontawesome.com/how-to-use/on-the-web/styling/stacking-icons

However, I decided not to use this approach, so I'll close this issue.

PR #64, opened 2020-09-16
Feat: Reverse page titles (argoproj/argo-cd#4337)

jsoref:


src/components/page/page.tsx
```diff
@@ -30,9 +30,9 @@ export const Page = (props: PageProps) => (
                         {(ctx) => {
                             const titleParts = [ctx.title];
                             if (toolbar && toolbar.breadcrumbs && toolbar.breadcrumbs.length > 0) {
-                                titleParts.push(toolbar.breadcrumbs.map((item) => item.title).join(' / '));
+                                titleParts = [toolbar.breadcrumbs.map((item) => item.title).reverse().join(' / ')].concat(titleParts);
```
alexmt, line 33: CI failed because titleParts is `const` and cannot be reassigned (https://app.circleci.com/pipelines/github/argoproj/argo-ui/89/workflows/c12a894f-2f43-4fab-a8b8-07e4d3e046b6/jobs/88):

```
TS2540: Cannot assign to 'titleParts' because it is a constant or a read-only property.
```

I think changing `const titleParts = [ctx.title];` to `let titleParts = [ctx.title];` should be enough to fix it.

alexmt (changes requested):

alexmt (approved):

PR #65, opened 2020-09-21
fix(ui): use hh:mm:ss notation (argoproj/argo-cd#4388)

jsoref:
Signed-off-by: Josh Soref <jsoref@users.noreply.github.com>

closes argoproj/argo-cd#4388


jsoref:
frown?

agilgur5:
I think this should be effectively superseded by #535

PR #68, opened 2020-10-10
refactor: enable strict mode in typescript

dcherman:
This pull request corrects all current typing issues present in argo-ui, then finally enables strict type checking.  While this PR is rather large, each commit should address a single file and is individually reviewable.  As much as possible, I avoided changing multiple components in the same commit to keep this reviewable.

The changes were made to err on the side of safety and backwards compatibility when introducing externally facing changes to the types.  That means that some of our APIs may not be ideal in this PR, however they reflect the actual ways in which the API could be used today; for some of the weirder cases, we should consider checking and refactoring argo/argo-cd (the only two consumers I think?) and then simplifying the types that we're exposing.

The lack of a test suite makes ensuring that all of these changes work with both argo workflows and argo cd challenging, so I am going to build both of those locally and do some basic manual testing with them, though we should absolutely add some actual unit/integration tests in the future to avoid needing to do that.

@simster7


simster7:
Sorry, I was out for a while and missed this PR. My main question: does this PR only contain changes to enable strict mode or does it also contain any functional changes?

dcherman:
@simster7 No worries.  The answer is that as best I could, I made changes only to support strict mode and not make functional changes.  If I had to identify the change I would be most concerned about making functional changes, it would likely be https://github.com/argoproj/argo-ui/pull/68/commits/d3b6bc667219965f383569ea0cbb84f61f8a8750

The lack of unit tests does make it challenging to ensure that no changes were made, so it's still a TODO for me to pull down both Argo Workflows/CD, build the UI with this PR, and click around a bit/do some manual testing.  I'm not planning to make that an exhaustive test, so it's still no guarantee.  See https://github.com/argoproj/argo-ui/pull/70 where I've started implementing unit tests which should hopefully make changes like this less risky in the future.

I am also more than happy to break this PR into many PRs to make them easy to review individually, then slowly land this over the course of a couple weeks (finally implementing strict mode at the end).

simster7:
Sounds good, let me discuss with the team to see what their thoughts are. I am not too opposed to just merging this as is.

However, if it's not too much work for you, you can break this into two PRs, one that is only minor/non-functional changes to appease strict mode (we can merge this quickly), and the other with all the more risky changes (we can make a more thorough review on that one). This way we can get most of the changes in and still have time to review the more significant ones.

simster7:
Thank you so much for doing this, too!

dcherman:
Sounds good.  I can likely work on that tonight a bit.  Once the testing PR has been merged, I can also introduce targeted unit tests for the riskier changes (before making them of course), to help ensure that no functional changes were made.

simster7:
@dcherman Any updates on this?

dcherman:
@simster7 I'll be making more PRs in the coming days related to this.  Should be able to do significantly more after this week hopefully, off for the rest of the year :)

PR #69, opened 2020-10-20
test: Add QE ID's to support test tool locators

keithchong:
The change adds properties and attributes only. This _should_ be enough to support at least the scenario described in https://github.com/argoproj/argo-cd/issues/4393.

Feel free to suggest other names of the IDs.  

This issue will fix: https://github.com/argoproj/argo-cd/issues/4598.  See issue for the reason why the `id` attribute is not used.

An example use case in the test code is to locate an input 'text' field with a specific QE ID:

`By.xpath(".//input[@qe-id='createAppPage-clusterName']"`

All defined identifiers can be placed in one file for easy maintenance.


src/components/autocomplete/autocomplete-field.tsx
```diff
@@ -24,6 +24,7 @@ export const AutocompleteField = ReactForm.FormField((props: AutocompleteProps &
             renderInput={(inputProps) => (
                 <input
                     {...inputProps}
+                    qe-id={props.qeId}
```
alexec, line 27: minor - I'm pretty sure this needs to be camel-case.

keithchong: Hi @alexec, this is a 'custom' attribute for the `input` element. I tried camel-case but it's an error because the property does not exist on type 'DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>'.ts(2322).

BTW. I had to make some additional changes.  I added ids to the OK and Cancel buttons in popup-manager.tsx.  (So that we can delete the app and dismiss the confirmation popup).

One other change is to the `qeId` property in autocomplete.tsx, which is now `lowercase` qeid.  This property is referenced in the FormComponent element in form-field.tsx.  In that case `qeid` is passed as a property to the AutoCompleteField component (autocomplete-field.tsx) but it _also_ can be an attribute to a React component, such as **Text**.  This is the only way I can get the form-fields to work so that there are no React errors or warnings.

keithchong:
Hi @alexec, here is an example of how this change will be used.

https://github.com/keithchong/argo-cd/commit/2da52b74a3e06c6fe5dd1bd9e607beb9cb4ed82d

See also the comments from https://github.com/argoproj/argo-cd/issues/4393

alexmt (approved):
LGTM

PR #71, opened 2020-11-05
refactor: correct typing errors in several components and dependencies

dcherman:
This is intended to be an easily reviewable subset of the changes proposed in #68


alexec (approved):
LGTM

PR #70, opened 2020-11-04
chore: introduce jest for unit testing

dcherman:
This introduces Jest as a framework for unit testing, including some
basic tests for the Checkbox.  The component being tested here is
intentionally minimal since this serves as our PR to set up the infra
for unit testing rather than introducing large amounts of tests.

PR #72, opened 2020-11-05
refactor: correct typing errors in several components and dependencies

dcherman:
This is intended to be an easily reviewable subset of the changes proposed in #68


package.json
```diff
@@ -25,6 +25,7 @@
         "foundation-sites": "^6.4.3",
         "handlebars": "^4.7.6",
         "history": "^4.7.2",
+        "is-promise": "^4.0.0",
```
alexec, line 28: accidentally added?

dcherman: Nope, intentional.  If we want to avoid the extra dependency, I'm happy to copy it rather than adding it as a dependency since it's only a couple of lines

alexec: Can I ask we try and minimise the argo-ui dependencies only to things is needs? argo-ui dependencies are 80% of Snyk the security alerts we get - so we create work for ourselves.

dcherman: Fixed.  Once https://github.com/argoproj/argo-ui/pull/70 has been reviewed/merged, I'll go back and add some unit tests for the new `isPromise` function.

alexmt (approved):
LGTM!

keithchong:
There is a missing '?' mark in the changes made to data-loader.tsx

See https://github.com/argoproj/argo-cd/issues/4913

dcherman:
@keithchong I'll take a look at this either later today or tomorrow.  It's actually more complex than adding a `?` due to the signature of `load` in `LoaderProps`

https://github.com/argoproj/argo-ui/blob/7b28e38b07f9e591cab784ac5ef2881bab4aaa8d/src/components/data-loader.tsx#L10

If you look at how argo-cd is passing `load`, it doesn't accept an optional argument, but rather accepts no arguments.  That means that even passing `undefined` would not pass the type checks.

The two fixes that I can see of the top of my head are:

1.  Fix the call in `argo-cd` to accept an optional state param, at which point we might as well just add `input` there.
2. Overload the props/state interfaces so that if `input` is omitted, your `load` function must accept no arguments and vice versa.  This is probably the route that I'll look into since it requires no changes in downstream consumers which was the intent of these changes - we can look into refactoring those projects later.

I also noted that no projects use the `noLoaderOnInputChange` prop as far as I can see which means it's a candidate for removal.  That'll be done separately since I want to triple check that assertion.

PR #73, opened 2020-11-06
fix(ui): Favicon should be circular w/ transparency (argoproj/argo-cd#4772)

jsoref:
closes argoproj/argo-cd#4772


alexmt (approved):
LGTM

PR #74, opened 2020-11-20
fix: Design Flaw leading to errant delete #4844

keithchong:
Signed-off-by: Keith Chong <kykchong@redhat.com>

This is the argo-ui part of the fix for https://github.com/argoproj/argo-cd/issues/4844

This 'general' fix improves the prompt window so clients can customize the header colour and icon colour.  Specifically,

- A new warning icon was added (ran `yarn utils:icons`)
- The layout of the popup prompt contents was improved when an icon is shown or not (the footer buttons should align with the main body contents)
- It allows for red, yellow title colours in the prompt window (Clients should put ‘red’ or ‘yellow’ or ‘normal’ values only for now)
- New property/option for title colour and icons have been added and are made optional to not break existing code and behaviour.
- Layout of `<h4>` and `<p>` elements improved.  All components should be aligned left with or without icon. Any custom indentation should be added by clients.
- Added five new storybook items as examples.

Note that the implementation of the icon colour was not changed (The use of `warning`, `success`, and `failed` was left unchanged, to not break existing code and behaviour)

**_Note_** that this change might require clients to modify their horizontal spacing (likely shrink or remove them) so that the components will better align with the footer buttons.

PR #75, opened 2020-11-26
fix: Design Flaw leading to errant delete - remove quote (#4844)

keithchong:

Remove unneeded quote.  This is continued from

https://github.com/argoproj/argo-ui/pull/74

which is the argo-ui part of the fix for argoproj/argo-cd#4844


keithchong:
Hi @alexmt, could please review and approve this?

alexmt (approved):
I should have noticed during review . Thank you for fixing!

PR #76, opened 2020-11-26
fix(data-loader): correctly support usage that excludes input entirely

dcherman:
https://github.com/argoproj/argo-cd/issues/4913


alexmt (approved):
LGTM

PR #81, opened 2020-12-07
feat: Allow for default values to be set in popup forms (#3205)

keithchong:

In general, this is to allow any popup form content to have default values pre-set by clients.  See the storybook entry.  Here, the username field default value is set to `admin` and the checkbox is checked by default, when the popup appears.  See screenshot:

Screen Shot 2020-12-07 at 3 05 42 PM

Specifically, this is needed by https://github.com/argoproj/argo-cd/issues/3205.  The confirmation dialog will be changed to a prompt, with confirmation to delete by requiring users to enter the application name.  That way, the Cascade checkbox can remain checked by default.

(Commit message has wrong issue.  It should be https://github.com/argoproj/argo-cd/issues/3205 and not 4844)

### DIFF
--- a/src/components/popup/popup-manager.tsx
+++ b/src/components/popup/popup-manager.tsx
@@ -14,6 +14,7 @@ export interface PopupApi {
         },
         customIcon?: {name: string, color: string},
         titleColor?: string,
+        defaultValues?: {},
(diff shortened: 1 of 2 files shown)


alexmt (approved):
LGTM!

PR #82, opened 2020-12-11
fix(autocomplete): correct typing errors

dcherman:
Signed-off-by: Daniel Herman <dherman@factset.com>


dcherman:
@simster7 Ping - any chance we can take a look at this?  We're very, very close to being done with this effort in argo-ui, then I'm planning to move on to doing the same thing in Argo Workflows and ArgoCD.

PR #85, opened 2020-12-15
fix(dropdown): correct typing errors

dcherman:
Signed-off-by: Daniel Herman <dherman@factset.com>

### DIFF
--- a/src/components/dropdown/dropdown.tsx
+++ b/src/components/dropdown/dropdown.tsx
@@ -1,7 +1,7 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import { BehaviorSubject, Observable, Subscription } from 'rxjs';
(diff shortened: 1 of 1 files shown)


src/components/dropdown/dropdown.tsx
```diff
@@ -18,12 +18,12 @@ export interface DropDownState {
 
 require('./dropdown.scss');
 
-const dropDownOpened = new BehaviorSubject<DropDown>(null);
+const dropDownOpened = new ReplaySubject<DropDown>(1);
```
dcherman, line 21: Both this and the old code have a small memory issue which I'll file an issue for separately.  Basically, even after the dropdown has been removed by React, this subject will end up retaining it in memory until a new dropdown is opened.

dcherman:
@simster7 Ping - any chance we can take a look at this?  We're very, very close to being done with this effort in argo-ui, then I'm planning to move on to doing the same thing in Argo Workflows and ArgoCD.

PR #88, opened 2021-01-23
fix: support Observable that return no data

alexmt:

PR fixes data loader bug. Observable might complete without returning any data. In this case, data loader should not stuck


rbreeze (approved):
LGTM

PR #90, opened 2021-01-26
feat: support programmatically control nav bar style

alexmt:

https://user-images.githubusercontent.com/426437/105889850-7cae6f00-5fc3-11eb-807f-a4452ffe0fa8.mov

PR allows to programmatically enable compact and light style of nav bar using `<Layout navItems={navItems} navBarStyle={{light: true, compact: true}}>`

### DIFF
--- a/src/components/layout/layout.tsx
+++ b/src/components/layout/layout.tsx
@@ -1,16 +1,18 @@
 import * as React from 'react';
-import { NavBar } from '../nav-bar/nav-bar';
+import { NavBar, Style } from '../nav-bar/nav-bar';
 
 require('./layout.scss');
 
-export interface LayoutProps extends React.Props<any> {
+export interface LayoutProps {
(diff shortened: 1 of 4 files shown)


alexec:
hmm... toggle styling .. I think we should just have one styling for simplicity

could compact be auto based on number of icons?

alexmt:
I like that idea. What about dark/light? I want to try to avoid changing color in Argo CD .

rbreeze:
I agree with @alexec to keep things simple, so maybe we should just revert the color to the old way, and change the icon size only if `numIcons >= 10`. 

The main reason I changed the color was for a simpler look (and simpler CSS) anyway - we should discard it for now, and could always revisit when CD has a major UI update if we wanted.

alexmt:
Sounds good. Implemented suggested change: nav bar automatically switches to "compact mode" if number of icons >= 10 and color change is reverted.

rbreeze (approved):
LGTM

PR #91, opened 2021-02-03
fix: autocomplete filter to make it case insensitive #5152

ishitasequeira:
For argoproj/argo-cd#[5152](https://github.com/argoproj/argo-cd/issues/5152)

Autocomplete filter has been made case insensitive

Signed-off-by: Ishita Sequeira isequeir@redhat.com


alexmt (approved):
LGTM!

You will need to run `yarn add https://github.com/argoproj/argo-ui.git` in `argo-cd/ui` to upgrade argocd to latest argo-ui version

PR #92, opened 2021-02-18
fix: data loader should return data if input has changed before observable returned first even

alexmt:

Another DataLoader bug fix: currently if data loader input changes before observable returns something it keep waiting forever. To reproduce:

* navigate to log viewer (e.g. https://cd.apps.argoproj.io/applications/argo-cd?node=%2FPod%2Fargocd%2Fargocd-redis-ha-haproxy-69c6df79c6-c9cfb%2F0&tab=logs) and enter a filter that does not match anything. 
* remove filter - loader keep waiting for results from previous request forewer


rbreeze (approved):
LGTM!

PR #93, opened 2021-03-12
fix: provide more visual cues for focused elements

reginapizza:
This change would introduce visual cues when focus is applied to buttons and checkboxes.

Buttons when applied focus would behave the same way as when a user hovers over them and will also have a 1px teal border around them. (OK is focused below)

Screenshot from 2021-03-12 15-26-45

Checkboxes would have a 2px teal outline when focused. 

Screenshot from 2021-03-12 15-25-43

### DIFF
--- a/src/styles/elements/buttons.scss
+++ b/src/styles/elements/buttons.scss
@@ -22,10 +22,18 @@
         &:hover {
             color: $argo-color-gray-6;
         }
+        
+        &:focus {
+            color: $argo-color-gray-6;
+        }
     }
 
     &:not(.disabled):not([disabled]) {
         &:hover {
+            background-color: #d1d5d9;
+        }
+
+        &:focus {
             background-color: #d1d5d9;
         }
     }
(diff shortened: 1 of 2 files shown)


src/styles/elements/buttons.scss
```diff
@@ -49,6 +63,11 @@
             &:hover {
                 color: $argo-color-gray-7;
             }
+
+            &:focus {
+                color: $argo-color-gray-7;
+                border: 1px solid $argo-color-teal-5 !important;
```
rbreeze, line 69: Does this style not work without the `!important`? I like to avoid using `!important` if possible

reginapizza: Trying to test it and it's actually not working with or without the `!important`. I think that disabled buttons are always skipped over in the DOM and never get focus. I've removed this entire pseudoclass.

rbreeze (approved):
LGTM

alexmt (approved):
LGTM

PR #98, opened 2021-04-07
feat: v1.1 (Live Explorer, all new hook-based components, documentation)

rbreeze:
## Proposal
Argo UI v1.1 replaces Storybook with a live component viewer. Users can play with Argo UI JSX code in their browser and see the results live, as well as documentation for components they are using. 

All new components are written using the more future-proof React Hooks. These new components are placed in a v2 directory so that they can coexist with older components for v1.1. In Argo UI v2, these legacy components would be removed. 

Some v2 components are direct replacements for v1 components, and others are brand new. v2 Components also introduce dark mode compatibility.

Finally, this PR introduces a new documentation paradigm that inlines documentation for each component directly into its code. This means: 

A) It's easy to find documentation since it's in the same file as the code it refers to
B) It's easy to format it however you wish
C) It's easier to remember to update documentation when you make code changes

## Live Explorer Demo:

https://youtu.be/CSYv7WOVWgg

Screen Shot 2021-04-07 at 3 25 15 PM


agilgur5:
Per https://github.com/argoproj/argo-ui/pull/100#issuecomment-828767541, this was closed as a Storybook replacement was deemed to onerous to maintain.

Also, correct me if I'm wrong, but nowadays doesn't Storybook support all these features? It may require plugins in some cases though

PR #99, opened 2021-04-14
docs: fix image

jef:
Fixes image and sets a max height of 200px so it's not overbearing.


alexec (approved):

alexec (changes requested):
please fix conflict

jef:
Looks like the link got updated. Thanks!

alexec (approved):

PR #113, opened 2021-07-02
fix: Change direction of autocomplete dropdown if there is no room fo…

rbreeze:


alexmt (approved):
LGTM

PR #115, opened 2021-07-16
fix: Adjust spacing of multiple info items in an info item row

rbreeze:


alexmt (approved):
LGTM

PR #116, opened 2021-07-16
Fix: prevent crash when Checkbox onChange is null

rbreeze:


alexmt (approved):
LGTM

PR #119, opened 2021-08-02
fix: Hide autocomplete suggestions on Enter but show on typing

rbreeze:


alexmt (approved):
LGTM!

PR #122, opened 2021-08-10
fix: autocomplete shuold be possible to use inside 'small' scrollable…

alexmt:

It should be possible to add autocomplete component with parent which is "smaller" than dropdown menu:

Before fix:

image

After fix:

image


rbreeze (approved):
LGTM! Thanks for catching `nodemon`. Must have added it on accident during dev

PR #127, opened 2021-08-13
feat: Allow CheckboxRows to be styled inline

rbreeze:


agilgur5:
This was component was deleted in #128

PR #123, opened 2021-08-10
fix: autocomplete items should be hidden using display: none

alexmt:
Signed-off-by: Alexander Matyushentsev <AMatyushentsev@gmail.com>

PR fixes minor visual glitch on safari:

image

PR #131, opened 2021-08-16
fix: autocomplete should expand into right direction

alexmt:

PR https://github.com/argoproj/argo-ui/pull/123 fixed visual glitch in safari but introduced another bug: autocomplete unable to expand into right directory because hidden content has no height. PR fixes the glitch in another way.


rbreeze (approved):
LGTM

PR #134, opened 2021-08-20
chore: Move date helper functions from Argo Workflows to Argo UI

rbreeze:
To help fix Argo CD issue: https://github.com/argoproj/argo-cd/issues/6924


v2/utils/utils.tsx
```diff
@@ -89,3 +89,55 @@ export const debounce = (fxn: () => any, ms: number) => {
         }, ms);
     };
 };
+
+export function formatDuration(seconds: number, sigfigs = 1) {
```
alexec, line 93: I think there were tests?

rbreeze: I couldn't find any tests for this, and I don't believe we have Jest set up in Argo UI. I could do this but I think it would be better in a separate PR

alexmt (approved):

PR #142, opened 2021-09-08
fix: logviewer long url incorrect line wrap when copying

tczhao:
fixes https://github.com/argoproj/argo-workflows/issues/6292

xterm version 2.9 fixes the issue.

There's an even newer version v3/v4+, in these versions,
they modified the Terminal interface, instead of taking in style class, they use js variable to define colours,
I couldn't find a way to make it reuse the common argo colour from css file without redefining them in js, thus didn't go with these versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
         "react-toastify": "^5.0.1",
         "rxjs": "^6.6.6",
         "typescript": "^4.0.3",
-        "xterm": "2.4.0"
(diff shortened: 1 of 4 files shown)


tczhao:
Hi @alexec, could you take a look at this PR please?

tczhao:
seems there will be compatibility issues with multiple project, close for now

PR #155, opened 2021-10-01
fix: Application details view performance issue #7262

oleksandr-codefresh:
Fixed endless re-rendering of Tabs component that was detected during investigation of https://github.com/argoproj/argo-cd/issues/7262


alexmt (approved):
LGTM

PR #156, opened 2021-10-01
fix: Cmd+Click on multiple applications opens each application in the same browser tab #7192

oleksandr-codefresh:
[NavigationManager]: fixed method that opens new tab when metaKey === true


pasha-codefresh:
https://github.com/argoproj/argo-cd/issues/7192

alexmt (approved):
Thank you!

PR #163, opened 2021-11-01
fix: change prop type of abbreviations to Map

saumeya:
Abbreviation prop type changed to reuse resources Map<> in argo-cd as per comment on https://github.com/argoproj/argo-cd/pull/7466


alexmt (approved):
LGTM

PR #178, opened 2022-02-10
fix(select): vertically center arrow

tczhao:
fixes https://github.com/argoproj/argo-workflows/issues/7789

center align select arrow, fixes hidden arrow on no text selection

left(old) -> right(new)
argo-workflow:
argo-workflow

argo-cd:
argocd

### DIFF
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -24,7 +24,7 @@
     }
     
     &__value-arrow {
-        position: absolute;
+        // right most, vertically centered
         top: 50%;
         right: 0;
         transform: translateY(-50%);
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -25,8 +25,9 @@
     
     &__value-arrow {
         position: absolute;
-        top: 8px;
+        top: 50%;
         right: 0;
+        transform: translateY(-50%);
     }
     
     &__options {


terrytangyuan (commented):
Thank you! I'll let others to double check as I am not familiar with this repo.

agilgur5:
@tczhao does the Workflows UI still have this issue? If so we can re-open and merge this

src/components/select/select.scss
```diff
@@ -25,8 +25,9 @@
     
     &__value-arrow {
         position: absolute;
-        top: 8px;
+        top: 50%;
         right: 0;
+        transform: translateY(-50%);
```
agilgur5, line 30: Confirmed this is [supported on all modern browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translateY#browser_compatibility) (I believe this didn't work on some older browsers back in the day so wanted to double-check)

agilgur5: Also could this be simplified to `top: 37.5%`?

tczhao:
Hi @agilgur5 ,

Thanks for looking into this,
we still have this issue,
image

Let me test with the suggested changes

tczhao:
We have the following
```
        position: absolute;
        top: 50%;
        right: 0;
        transform: translateY(-50%);
```
image

---

```
        position: absolute;
        top: 37%;
        right: 0;
```
image

---

```
        position: absolute;
        top: 0;
        right: 0;
```
image

agilgur5:
Oh boy, on my monitor, it turns out I've never even seen this arrow, which is probably why I was a little confused when I first looked at your screenshot.

<details> <summary> <code>top: 8px;</code> -- the original. invisible/below the fold </summary>

top 8px original - Screenshot 2024-06-10 at 9 16 36 PM

</details>

<details> <summary> <code>top: 0;</code> -- too low </summary>

top 0 - Screenshot 2024-06-10 at 9 13 17 PM

</details>

<details> <summary> <code>top: 50%; transform: translateY(-50%); </code> -- just right </summary>

top 50 plus translate -50 -- Screenshot 2024-06-10 at 9 15 21 PM

</details>

src/components/select/select.scss
agilgur5: The answer is no, see [below comments](https://github.com/argoproj/argo-ui/pull/178#issuecomment-2157301364)

agilgur5:
Gosh, my CSS is definitely a bit rusty, this is vertical centering, right, which is one of the most hacky CSS things. `top: 50%; transform: translateY(-50%);` means half-way down, and then half of the element's _own_ height back up -- see also this [SO answer](https://stackoverflow.com/a/40530404/3431180)

src/components/select/select.scss
```diff
@@ -25,8 +25,9 @@
     
     &__value-arrow {
         position: absolute;
```
agilgur5, line 27:
```suggestion
        // right most, vertically centered
```

add a comment for clarity. also we need DCO to re-run 😉

agilgur5 (approved):
Thanks for fixing and double-checking!

src/components/select/select.scss
agilgur5: Oops I made a mistake here and accidentally removed the `position: absolute;` 😬 
Fixed in #563

PR #179, opened 2022-02-11
feat: add optional prop closeOnEscape to sliding panel

reginapizza:
This and [Argo CD PR#8465](https://github.com/argoproj/argo-cd/pull/8465) would close [#8225](https://github.com/argoproj/argo-cd/issues/8225) in argocd. 

Adds additional optional prop `closeOnEscape` to the sliding panel, making closing sliding panels as easy as pressing the escape key.

### DIFF
--- a/src/components/sliding-panel/sliding-panel.tsx
+++ b/src/components/sliding-panel/sliding-panel.tsx
@@ -8,6 +8,7 @@ export interface SlidingPanelProps extends React.Props<any> {
     hasNoPadding?: boolean;
     offCanvas?: boolean;
     hasCloseButton?: boolean;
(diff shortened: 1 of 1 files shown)


rbreeze (approved):
LGTM, and I'm okay with merging as is, but did you consider using the `KeybindingProvider` in https://github.com/argoproj/argo-ui/blob/master/v2/shared/keypress.tsx ?

crenshaw-dev:
@reginapizza was there any follow-up here, or should we merge?

wtam2018:
This is no longer needed because we implemented the key binder solution https://github.com/argoproj/argo-ui/pull/184

PR #184, opened 2022-03-29
feat: Escape key should close sliding window in UI (#8225)

keithchong:
KeybindingProvider solution

Key Points:

- This may affect existing clients of the Sliding Panel since this solution includes wrapping a KeyBindingProvider. We could remove this (or make it an option) and simply enforce the provider to be added in the client code instead
- KeyBindingProvider can be removed on clients of the SlidingPanel
- KeyBindingProvider therefore needs to be adjusted in Argo CD's ApplicationsList page (minor change to simply move that element to another location).  However, it works even without the change.
- Focus is on the close button if present whenever the sliding panel opens. That way, users can press escape button to close it right away, especially when the panel is 'read-only'.
- Sliding panel now removes close button when hidden but adds it back when reappearing so that focus is in effect
- This adds support for multiple targets for the provider. There can be multiple targets for the sliding panel. The current 'target' can be removed if desired.
- Note that this solution generally does not close the slider panel every time the escape key is pressed. It is correct to do this because it depends on the current active element at the time the key is pressed. (There can be an editable panel in the slider, and a text field can be active, and if you press escape, the panel should not close)
- There is strange behaviour when pressing escape on a (closed) combo box and it closes the slider, even though the panel is still in edit model.  It appears to have lost focus. So this is still an existing problem.  This focus loss problem does not occur on text fields or buttons
- Tested this on sliding panels in applications list page (the main page), as well as the details page, and most settings pages. All of these do not require changes except the main applications list page (see bullet point 3 above)

https://github.com/argoproj/argo-cd/issues/8225


reginapizza (approved):

keithchong:
Hi @rbreeze, could you please review this? Thanks in advance.

v2/shared/keypress.tsx
```diff
@@ -104,6 +104,7 @@ export interface KeyFxnProps {
     action: KeyAction;
     combo?: boolean;
     target?: React.MutableRefObject<any>;
+    targets?: React.MutableRefObject<any>[];
```
rbreeze, line 107: I prefer to combine the `target` and `targets` props into one. To maintain backwards compatibility, it can retain the name `target`, but be of type `React.MutableRefObject<any> || React.MutableRefObject<any>[]`. Then we can  check if this prop is an array with `if (target.length)`

src/components/sliding-panel/sliding-panel.tsx
```diff
@@ -15,36 +16,78 @@ export interface SlidingPanelProps extends React.Props<any> {
...
+    const panelHeaderDivRef = React.useRef(null);
+    const panelFooterDivRef = React.useRef(null);
+
+    React.useEffect(() => {
+        if (closeButtonRef && closeButtonRef.current) {
+            closeButtonRef.current.focus();
+        }
+    }, []);
+
+    if (props.isShown) {
```
rbreeze, line 42: We shouldn't call a hook inside an if condition (see https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

Instead, you can move this condition to inside the `action`

rbreeze (changes requested):

src/components/sliding-panel/sliding-panel.tsx
keithchong: Hey @rbreeze , thanks for catching this. I'll move it to the existing if condition inside.

v2/shared/keypress.tsx
keithchong: Hi @rbreeze , I've combined them.  keypress.tsx was updated to combine them.  See the bulk of the changes starting from line 132 to handle both situations.   I tested this with pressing escape on any opened sliding panel, as well as pressing the forward slash on the application list page to activate the search bar.

src/components/sliding-panel/sliding-panel.tsx
keithchong: Done

rbreeze (approved):

Issue #221, opened 2022-05-20
Accessibility: side navbar


gaurav5430:
These are some of the accessibility issues i have identified on the side nav bar, with some quick solutions (may not be the best):

- there is no indication that this is a menu or nav - can use `nav` element
- there is no indication for the current active item for the screenreader user - can use `aria-current="true"`
- keyboard users can focus on the next item using tab, but can't activate it using space / Enter - can use `button` instead of a `div` with onClick
- there is a visual change on mouse hover but not on keyboard focus (although the tooltip gets shown, which is an indication that the focus has changed) - apply the same styles as hover

saumeya:
@gaurav5430 can you share screenshots of the change, thanks!

Issue #166, opened 2021-12-11
nav bar and tab don't have a tag link histories


krrrr38:
argo-cd and argo-workflows Nav bar doesn't have `a` tag. So currently we cannot open it from `right click`.

In addition, browser doesn't register page histories, so we cannot `go forward` history after `go back` page.

![image](https://user-images.githubusercontent.com/915731/145677592-f7d2fd42-4e9a-4a57-bd99-e3bfbafbe174.png)

Generally if it has `a` tag, we can open it from `right click` like followings.

![image](https://user-images.githubusercontent.com/915731/145677624-9ed48591-93f1-4ae1-852c-c4de01e1912e.png)

----

I'm not sure this is `argo-ui` issue or not.

PR #222, opened 2022-05-20
fix(side-navbar): keyboard a11y

gaurav5430:
replace div with button to allow activation using space and enter.

This fixes one of the issues mentioned in https://github.com/argoproj/argo-ui/issues/221

### DIFF
--- a/src/components/nav-bar/nav-bar.tsx
+++ b/src/components/nav-bar/nav-bar.tsx
@@ -35,10 +35,10 @@ export const NavBar: React.FunctionComponent<NavBarProps> = (props: NavBarProps,
                 <div className='nav-bar__version'>{props.version && props.version()}</div>
                 {(props.items || []).map((item) => (
                     <Tooltip content={item.title} placement='right' arrow={true} key={item.path + item.title}>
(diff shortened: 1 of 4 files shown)


agilgur5:
Looks like DCO is missing a space (between name and the `<`)

gaurav5430:
> Looks like DCO is missing a space (between name and the `<`)

Thanks for reopening this, I might find time to fix the DCO issue and rebase this

src/components/nav-bar/nav-bar.tsx
```diff
@@ -35,10 +35,10 @@ export const NavBar: React.FunctionComponent<NavBarProps> = (props: NavBarProps,
                 <div className='nav-bar__version'>{props.version && props.version()}</div>
                 {(props.items || []).map((item) => (
                     <Tooltip content={item.title} placement='right' arrow={true} key={item.path + item.title}>
-                        <div className={classNames('nav-bar__item', { active: isActiveRoute(locationPath, item.path) })}
+                        <button type="button" className={classNames('nav-bar__item', { active: isActiveRoute(locationPath, item.path) })}
```
agilgur5, line 38: I think this can actually just use a `<Link>` element from `react-router`, which would also solve #166

Issue #286, opened 2022-09-28
Helm icon is not being rendered correctly


tongpu:
I've always wondered that the Helm icon in the Argo CD UI looks a bit strange, but today I decided to go down that rabbit hole and found out that this is actually being rendered by custom icon font. This screenshot is from Edge, but it looks similarly bad on all other browser I had available (Chrome and Firefox):

![image](https://user-images.githubusercontent.com/2674075/192767554-1c4bd5b1-7ecd-40be-9090-7ab5ff9026ad.png)

agilgur5:
#287 had a follow-up with #363

Thanks for going down the rabbit hole and fixing this!

tongpu:
> Thanks for going down the rabbit hole and fixing this!

Happy to help. The typography nerd in me was quite happy that I was able to figure it out. :D

PR #287, opened 2022-09-28
Update Helm icon

tongpu:
As a first follow-up to #286 I looked into the different SVG files and I had the feeling that some with multiple paths are not translated into `argo-icons.svg` correctly.

This is an updated version of the black Helm icon taken from the [CNCF Branding](https://cncf-branding.netlify.app/projects/helm/) page with the paths combined into a single path.


tongpu:
@saumeya @rbreeze Would you be willing to help me get this sorted out?

saumeya (approved):
lgtm

saumeya:
closes #286

rbreeze (approved):

tongpu:
I think the font files in `src/assets/fonts` would also need to be updated, but that was the part that I didn't manage to figure out yet and hoped that you have some tooling which generates the font files from the directories of SVGs.

PR #337, opened 2023-02-24
Customize autoclose delay timer value

ashutosh16:
Enable users to set a custom delay timer value for the autoclose property parameter.

### DIFF
--- a/src/components/notifications/notifications.tsx
+++ b/src/components/notifications/notifications.tsx
@@ -15,7 +15,7 @@ export enum NotificationType {
 export interface NotificationInfo {
     type: NotificationType;
     content: React.ReactNode;
-    autoCloseTimer?: any;
+    autoCloseTimer?: number | boolean;
 }
 
 export interface NotificationsProps {
@@ -57,7 +57,7 @@ export class Notifications extends React.Component<NotificationsProps> {
                 pauseOnHover: true,
                 pauseOnFocusLoss: true,
                 draggable: false,
-                autoClose: next.autoCloseTimer ? next.autoCloseTimer : AUTO_CLOSE_TIMEOUT,
+                autoClose: next.autoCloseTimer === undefined ? AUTO_CLOSE_TIMEOUT :  next.autoCloseTimer,
             });
         });
     }
--- a/src/components/notifications/notifications.tsx
+++ b/src/components/notifications/notifications.tsx
@@ -15,6 +15,7 @@ export enum NotificationType {
 export interface NotificationInfo {
     type: NotificationType;
     content: React.ReactNode;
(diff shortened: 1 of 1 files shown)


ashutosh16:
@alexef  Would you be able to review the PR? The change allows developer to customizing the notification delay time.

alexef:
Why would this be a per-notification setting, and not a global one? When would you want one to stick / disappear differently than the rest?

ashutosh16:
> Why would this be a per-notification setting, and not a global one? When would you want one to stick / disappear differently than the rest?

Currently, the notification setting is set [global](https://github.com/argoproj/argo-ui/blob/4fe8b83e418a65336f69ff67d31eb6aa7eaf8167/src/components/notifications/notifications.tsx#L7) and applies to all notifications. However, I'm working on a new feature that automatically switches the UI to the group node when the number of pods in the tree view exceeds a certain threshold. I plan to send a notification to the user when this switch occurs, and I want the notification to stick for a longer time so that users can read it.  This feature should not affect the global notification setting, which will continue to function as it does now. However some error notifications are too quick to read, so I think adding this property should provide a way to further customize the notification.

ashutosh16:
> > Why would this be a per-notification setting, and not a global one? When would you want one to stick / disappear differently than the rest?
> 
> Currently, the notification setting is set [global](https://github.com/argoproj/argo-ui/blob/4fe8b83e418a65336f69ff67d31eb6aa7eaf8167/src/components/notifications/notifications.tsx#L7) and applies to all notifications. However, I'm working on a new feature that automatically switches the UI to the group node when the number of pods in the tree view exceeds a certain threshold. I plan to send a notification to the user when this switch occurs, and I want the notification to stick for a longer time so that users can read it. This feature should not affect the global notification setting, which will continue to function as it does now. However some error notifications are too quick to read, so I think adding this property should provide a way to further customize the notification.

Does this explanation sound reasonable to you, If yes, can you approve the PR?

alexef:
While the explanation make sense, I would also explore other options. What if instead of persisting the notification longer, we disable auto close completely?

In which case, one has to manually suppress it, same as with the top level banners.

otoh I get the intention is to get user’s attention, as the default view was changed, but sending a notification every time one opens up the application details view could soon become annoying, and the next step would be to implement some cookie storage to know if the notification was shown already.

i don’t want to block the pr, as the change is minor, but these are my thoughts

alexef (approved):

ashutosh16:
@alexef  I agree with you - When rendering notifications, the user has the option to set a default timeout or override the behavior by setting autoClose to false. If the user only wants to display the notification once, this can also be handled at the implementation level.

@crenshaw-dev  Would you be able to merge the PR?

PR #341, opened 2023-03-16
fix for argocd issue 12798

reginapizza:
Submitting as part of the fix for https://github.com/argoproj/argo-cd/issues/12798 

Reviewer(s) please double check that other places using AutocompleteField component are not negatively affected by this change. I checked the ones I found but I'm not 100% confident I didn't miss any.


ciiay:
LGTM
Working great 👍

PR #346, opened 2023-03-19
fix: gradient changes in chat now button

2022H1030042G:


agilgur5:

PR #347, opened 2023-03-24
Gradient changes

2022H1030042G:


src/styles/elements/buttons.scss
```diff
@@ -235,9 +235,8 @@
     }
 
     &--special {
-        color: #fff;
-        background: linear-gradient(-58deg, #49505c 0%, #3f51b5 36%, #00c7d6 100%);
-
+        color: #fff;       
+	background-color: #3f51bf;
```
saumeya, line 239: Could you use color from pre-defined style like so - 
`background-color:  $argo-color-gray-6`
Check the colors in colors.scss file.

saumeya (changes requested):

agilgur5:

PR #481, opened 2023-11-27
Feature/npm and yarn/react router dom 6.20.0

sunyeongchoi:
https://github.com/argoproj/argo-ui/pull/480

dependabot automatically updates the `react-router-dom` version, but as the version is upgraded 6, the functions used change, causing errors in the build.

So, I changed it to use a function that matches the react-router-dom 16 version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
         "react-autocomplete": "1.8.1",
         "react-form": "^2.16.0",
         "react-helmet": "^6.1.0",
-        "react-router-dom": "^4.2.2",
+        "react-router-dom": "^6.19.0",
         "react-toastify": "9.0.8",
         "rxjs": "^7.8.1",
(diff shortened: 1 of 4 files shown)


agilgur5:
Note that this a (very) breaking change, so it does affect all downstream UI builds as well which we have to be fairly careful with.

I took care of the last remaining `react-router` v4 piece in Workflows in https://github.com/argoproj/argo-workflows/pull/11891. I haven't checked all [v5 to v6 breaking changes](https://reactrouter.com/en/main/upgrading/v5) yet though. We'd also probably want to upgrade to v5.1 first, per [the upgrade guide](https://reactrouter.com/en/main/upgrading/v5#upgrade-to-react-router-v51)

Also not sure if CD is able to upgrade. If not possible yet, they'd have to stay on the version before this gets merged.

PR #534, opened 2024-02-12
chore(deps): remove unused deps & `resolutions`

agilgur5:
Originally started as a follow-up to https://github.com/argoproj/argo-workflows/pull/12097#discussion_r1375279098 on `moment-timezone`, but there were actually _much_ more unused deps to prune

#### Deps
- these are all completely unused in the current codebase (search codebase, no references)
  - maybe some of these were used at some point in the past, but none of them are currently used

#### `resolutions`

- `lodash` resolution is no longer needed as all deps are compatible with `lodash` `^4`
  - #141 added it and it _may_ have been necessary for some of the `lodash` v3 deps there, but this codebase no longer has any `lodash` v3 deps

- `@types/history` is already using `^4.7.8` without the resolution
  - #71 added it while also upgrading the devDep; only upgrading the devDep was necessary

### Modifications

- remove unused `moment-timezone` from `dependencies`
- remove several unused `devDependencies`
- remove a few unnecessary or unused Yarn `resolutions`

### Verification

- installs, builds, tests, etc still work without all these

### Notes to Reviewers

I actually wrote this on Jan 6th (see the commits), the same time as my other PRs, but I thought this would merge conflict with some of them, so I didn't create the PR, intending to rebase with those once they were merged. This actually seems to not conflict with #511 at all as far as I can tell, so creating the PR now, and because it's already been over a month with no review/merge 😕  **EDIT**: Welp, they actually conflicted with #509. I have a feeling past me predicted that and then I just forgot over the past month


agilgur5:
Fixed merge conflicts with #509.

There's a new warning during the install that seems to be because dependabot just merged #539 and #541 (they were created before #537 was merged), which are almost certainly not compatible 😕 . They seem to have finally passed CI because #509 removed some old deps. That does bolster the case of #537

agilgur5:
`lint` is weirdly [erroring](https://github.com/argoproj/argo-ui/actions/runs/7889119339/job/21528202750?pr=534) here but not in other PRs that similarly don't change source code... strange...

Raised #543 for that since it's independent of this PR. Will have to rebase once that's merged

agilgur5:
> `lint` is weirdly [erroring](https://github.com/argoproj/argo-ui/actions/runs/7889119339/job/21528202750?pr=534) here but not in other PRs that similarly don't change source code... strange...
> 
> Raised #543 for that since it's independent of this PR. Will have to rebase once that's merged

I may have figured out why that's happening, some of the ESLint deps ended up getting upgraded a few minors in this PR. Let me fix that

crenshaw-dev (approved):

Issue #469, opened 2023-11-03
Project setup fails with module build error 


surajyadav1108:
followed the setup guide for the setup of argo-ui with given steps 
- `yarn install`
- `yarn start`
    `yarn start` failed with 
```
[Error message "error:0308010C:digital envelope `routines::unsupported](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported)
```

using `export NODE_OPTIONS=--openssl-legacy-provider` worked to start the build and server  but on server module build fails with 
```
Module build failed (from ./node_modules/ts-loader/index.js):
TypeError: loaderContext.getOptions is not a function
    at getLoaderOptions (/workspace/argo-ui/node_modules/ts-loader/dist/index.js:91:41)
    at Object.loader (/workspace/argo-ui/node_modules/ts-loader/dist/index.js:14:21)
```

agilgur5:
Yes you'll need to use the version of Node specified in the `.nvmrc`.

This repo is on an older/EoL version of Node (v14) and the OpenSSL error in later versions of Node is due to an incompatibility in Storybook v6/Webpack v4 (which is also outdated). I'm attempting to fix that in https://github.com/argoproj/argo-ui/pull/432#issuecomment-1879816619 but right now you should use the version of Node specified in the repo.

Also I'm not sure what work you were trying to do here, but this repo is no longer actively maintained, hence some issues like this will crop up. See #453 for more details.

agilgur5:
Welp, I reproduced the second error as well after updating Node and adding `NODE_OPTIONS=--openssl-legacy-provider` to all the Storybook scripts.

This seems to be because we're using Storybook v6 w/ Webpack v4, but dependabot updated `typescript-loader` from v8 to v9 in #227. And `ts-loader` v9 [requires Webpack v5](https://github.com/TypeStrong/ts-loader/blob/v9.5.1/CHANGELOG.md#v900), hence the old, incompatible APIs.

I think CI didn't fail on #227's build b/c CI [only checks `yarn build`](https://github.com/argoproj/argo-ui/blob/c65a9520366b0c0cf0ac585618d029e60041a94d/.github/workflows/build-v1.yml#L18) and not `yarn start`. In `yarn build`, this error pops up as a warning and it still has exit code 0:
```
❯ yarn build
yarn run v1.22.19
$ NODE_OPTIONS='--openssl-legacy-provider' build-storybook -o ./dist/storybook
[...]

PR #536, opened 2024-02-12
chore(deps): upgrade to node v20

agilgur5:
Supersedes / Closes #432

### Motivation

- node v14 is EoL and has been [since April 2023](https://github.com/nodejs/Release/blob/68840fae8fdaa59bf9431661238dbb86b4390829/README.md?plain=1#L7)
- node v20 is the [current LTS version](https://github.com/nodejs/Release/blob/9cf9c6ed8178e88c74dfa1921a793701e1d162c4/README.md?plain=1#L8)

### Modifications

- update `.nvmrc` to Node v20
- ~update all `actions/setup-node` GH Actions to use the `.nvmrc` for consistency~ **EDIT**: that ended up being completed as part of #509 

- due to an old version of Storybook / Webpack, we have to run all Storybook scripts with `NODE_OPTIONS='--openssl-legacy-provider'`
  - this must be used if we use newer versions of node without upgrading Storybook + Webpack
    - unfortunately, upgrading Storybook is significantly easier said than done, especially since there are 3 different Storybook builds in this repo (v1, v2, and `antd`). See https://github.com/argoproj/argo-ui/pull/432#issuecomment-1879816619 noting my trials and tribulations with that
      - so put that off for now (or ever? since this repo is not maintained much anymore)
        - potentially leave Storybook as-is and have it use Webpack v5? See also https://github.com/argoproj/argo-ui/issues/469#issuecomment-1879870424
  - note that this is an insecure run-time configuration (which is why it was deprecated and is legacy), but since it is only used in development, it's not a significant vulnerability
  
### Verification

Install works, builds pass, tests pass

### Future Work

1. Storybook upgrade, maybe never. 
    - See https://github.com/argoproj/argo-ui/pull/432#issuecomment-1879816619 on how difficult that is
1. Webpack upgrade, maybe doable. Then can remove `--openssl-legacy-provider`
    - Storybook v6 [supports Webpack v5 optionally](https://release-6-5--storybook-frontpage.netlify.app/docs/6.5/builders/webpack#webpack-5) (Webpack v4 is the default)
1. Have dependabot stop updating deps with breaking changes... 😐 
    - **EDIT**: Completed in #537


agilgur5:
Fixed merge conflicts

.nvmrc
```diff
@@ -1 +1 @@
-v14
+v20
```
csantanapr, line 1: Just as heads up to record here, ArgoCD-UI uses nodejs v21 (Current not LTS)
I think is ok to leave it at version v20 much improvement jump from v14

csantanapr (approved):
LGTM

.nvmrc
agilgur5: LTS is generally recommended. Workflows is also on [v20](https://github.com/argoproj/argo-workflows/blob/a927379f761fbffe733857e7a23d69511fba90de/ui/package.json#L16).

As I wrote [in Slack](https://cloud-native.slack.com/archives/C020XM04CUW/p1708632758034819?thread_ts=1708631268.852329&cid=C020XM04CUW), it's a devtool for front-end in any case, so it does not impact downstream

isubasinghe (approved):

PR #543, opened 2024-02-14
fix(lint): replace non-existent `qe-id` with `data-qe-id`

agilgur5:
### Motivation

- `lint` was erroring out: `Unknown property 'qe-id' found  react/no-unknown-property`

I'm not sure why it didn't pop up during #509 though... It's also only popping up on certain PRs for some reason, but not all of them. Like it pops up on #534 but not #536, despite neither of them changing any source code

### Modifications

- `qe-id` -> `data-qe-id`
  - [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) are the standardized way of persisting custom data to HTML
    - `qe-id` was used for testing purposes in this repo - this type of test is no longer standard, particularly in React, but for now I just renamed it to get `lint` passing
    
### Verification

`lint` passes now

### DIFF
--- a/src/components/autocomplete/autocomplete-field.tsx
+++ b/src/components/autocomplete/autocomplete-field.tsx
@@ -24,7 +24,7 @@ export const AutocompleteField = ReactForm.FormField((props: AutocompleteProps &
             renderInput={(inputProps) => (
                 <input
                     {...inputProps}
-                    qe-id={props.qeid}
+                    data-qe-id={props.qeid}
                     onFocus={(e) => {
                         if (inputProps.onFocus) {
                             inputProps.onFocus(e);
--- a/src/components/dropdown-menu.tsx
+++ b/src/components/dropdown-menu.tsx
@@ -21,7 +21,7 @@ export class DropDownMenu extends React.PureComponent<DropDownMenuProps> {
         return (
             <DropDown anchor={this.props.anchor} isMenu={true} ref={(dropdown: any) => this.dropdown = dropdown} qeId={this.props.qeId}>
                 <ul>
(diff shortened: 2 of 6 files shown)


crenshaw-dev (changes requested):
Testing locally, it looks like this change breaks Argo CD's application list autocomplete. Here it is after reverting to HEAD. 
image

agilgur5:
Well that's surprising, `qe-id` isn't a valid attribute on `input` elements and is a convention for tests... It shouldn't be used for any actual behavior at all 🤔

agilgur5:
> Testing locally, it looks like this change breaks Argo CD's application list autocomplete.

Even more confusingly, [CD's Application List](https://github.com/argoproj/argo-cd/blob/24ac326384b7939fa36d48e79cfa320528cf85e2/ui/src/app/applications/components/applications-list/applications-list.tsx#L228) uses the [`Autocomplete`](https://github.com/argoproj/argo-ui/blob/17d7a8be6aa2c0357f40bbe6856a1b2130806cd6/src/components/autocomplete/autocomplete.tsx) component, which did not change here (not the `AutocompleteField` component, which did change here)

agilgur5:
> Like it pops up on #534 but not #536, despite neither of them changing any source code

Figured out that was due to an accidental ESLint upgrade in #534's `yarn.lock` (https://github.com/argoproj/argo-ui/pull/534#issuecomment-2106058288) which I then reverted and got that one passing and merged.

So this is no longer needed but may be if ESLint is upgraded. Kicking this back to draft due to the [above](https://github.com/argoproj/argo-ui/pull/543#pullrequestreview-1904759991) issue then.

PR #545, opened 2024-03-17
fix: The navItems are not clickable with vimium (accessibility)

michaelfresco:
## Motivation

- This PR adds a tag `role="checkbox"` to the navItems. This has the benefit that keyboard users can use vimium to navigate the user interface. At the moment, the navItems do not show up as links, because they do not contain a typical `<a href>`

- Adding `role="checkbox"` works around that issue, and makes them clickable

- Please also see the original issue: https://github.com/argoproj/argo-workflows/issues/12813

### DIFF
--- a/src/components/nav-bar/nav-bar.tsx
+++ b/src/components/nav-bar/nav-bar.tsx
@@ -35,10 +35,12 @@ export const NavBar: React.FunctionComponent<NavBarProps> = (props: NavBarProps,
                 <div className='nav-bar__version'>{props.version && props.version()}</div>
                 {(props.items || []).map((item) => (
                     <Tooltip content={item.title} placement='right' arrow={true} key={item.path + item.title}>
+                    <a href="#">
                         <div className={classNames('nav-bar__item', { active: isActiveRoute(locationPath, item.path) })}
                             onClick={() => context.router.history.push(item.path)}>
                             <i className={item.iconClassName}/>
                         </div>
+                    </a>
                     </Tooltip>
                 ))}
             </div>


src/components/nav-bar/nav-bar.tsx
```diff
@@ -36,7 +36,7 @@ export const NavBar: React.FunctionComponent<NavBarProps> = (props: NavBarProps,
                 {(props.items || []).map((item) => (
                     <Tooltip content={item.title} placement='right' arrow={true} key={item.path + item.title}>
                         <div className={classNames('nav-bar__item', { active: isActiveRoute(locationPath, item.path) })}
-                            onClick={() => context.router.history.push(item.path)}>
+                            onClick={() => context.router.history.push(item.path)} role="checkbox">
```
agilgur5, line 39: > * At the moment, the navItems do not show up as links, because they do not contain a typical `<a href>`

why not use an actual `<Link>` component here? Same as https://github.com/argoproj/argo-ui/pull/222#discussion_r1597553119, that would resolve #166 as well

tooptoop4:
where @michaelfresco

michaelfresco:
Hi @tooptoop4, I have changed the <div> to a <Link> component as suggested by @agilgur5.

src/components/nav-bar/nav-bar.tsx
michaelfresco: I tried a few variations with a `<Link>` component but I don't know how to make this work. (I have little experience with web development). However, I think that it is a Tooltip-thing. If you look at other Tooltip examples, they have the same issue. Like here, https://react-tooltip.com/docs/getting-started, the Tooltip buttons are clickable but do not show up as real links.

tool-tip-demo

tooptoop4 (approved):

michaelfresco:
This version is actually creating side effects. Please don't 📦 .

Issue #512, opened 2024-01-08
UI error notification doesn't fit


JessieTeng89:
Describe the bug
Error message popup cannot be displayed properly

To Reproduce

At project level, add new role
Grant "get" to all resources under the related app
Project --> edit --> save
Expected behavior
Error message should be displayed properly, at least all the text should be visible

![image](https://github.com/argoproj/argo-ui/assets/101035990/4785c6a9-dc67-4d04-9a20-fe86654a6d62)

Related to https://github.com/argoproj/argo-cd/issues/16745

JessieTeng89:
I will create a PR to fix it

JessieTeng89:
Hi @agilgur5
I am trying to create a PR for this issue, but got following error message, any idea?
![image](https://github.com/argoproj/argo-ui/assets/101035990/9c0cb330-d5e1-4942-b6f9-598fc4be33a9)

agilgur5:
I've never seen this error before, but here's a GH community discussion on it: https://github.com/orgs/community/discussions/61472

Did you properly fork this repository and create a branch on your fork? I didn't see an `argo-ui` fork in [your public repos](https://github.com/JessieTeng89?tab=repositories)

JessieTeng89:
> I've never seen this error before, but here's a GH community discussion on it: https://github.com/orgs/community/discussions/61472
> 
> Did you properly fork this repository and create a branch on your fork? I didn't see an `argo-ui` fork in [your public repos](https://github.com/JessieTeng89?tab=repositories)

Thanks, I have fixed that issue and I will create a PR soon

PR #546, opened 2024-04-05
fix(notification): fit long words

JessieTeng89:
Fixes https://github.com/argoproj/argo-ui/issues/512
It is a layout issue when there is a long word in the error message toast popup, so added a style in that component to fix the issue.

Screenshots:
before fix:
image

after fix:
image


src/components/notifications/notifications.tsx
```diff
@@ -57,6 +57,9 @@ export class Notifications extends React.Component<NotificationsProps> {
                 pauseOnFocusLoss: true,
                 draggable: false,
                 autoClose: AUTO_CLOSE_TIMEOUT,
+                style: {
+                    wordBreak: "break-all"
```
agilgur5, line 61:
```suggestion
                    overflowWrap: "break-word"
```

[`break-word`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) might be more legible. but test it out and make sure

agilgur5: [Looks like](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#comparing_overflow-wrap_word-break_and_hyphens) the main difference is that it will add a newline before the long word. I'm honestly not sure if that's more legible or not 😅

JessieTeng89: > [Looks like](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#comparing_overflow-wrap_word-break_and_hyphens) the main difference is that it will add a newline before the long word. I'm honestly not sure if that's more legible or not 😅
> 
> I think it would avoid the "denie-d" in your screenshot and would make it on one line though? That would be better for shorter words that _could_ fit on one line then

Thanks for your suggestion, you are correct, it is much better to use `break-word` instead of `break-all` so that it only breaks the long words.
But I found it is only working with `word-break: break-word;` and it is not working with `overflow-wrap: break-word`, I found someone is facing the same issue with this component, I can change my PR to `word-break: break-word;` if it is good to you.
https://github.com/fkhadra/react-toastify/issues/112

After using `word-break: break-word;`:
image

agilgur5: `word-break: break-word` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#browser_compatibility). It appears to still be supported, but could be removed from browsers, so is not necessarily stable. That's why I suggested `overflow-wrap: break-word`, which is supported.

Strange that it didn't work for you, I wonder if there's some other CSS needed on the `display` or `overflow`? 🤔 
Or if it was overridden by something in the existing CSS?

Another option is `overflow-x: scroll`, which is not the nicest, but works relatively well with touch pads at least.

JessieTeng89: > `word-break: break-word` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#browser_compatibility). It appears to still be supported, but could be removed from browsers, so is not necessarily stable. That's why I suggested `overflow-wrap: break-word`, which is supported.
> 
> Strange that it didn't work for you, I wonder if there's some other CSS needed on the `display` or `overflow`? 🤔 Or if it was overridden by something in the existing CSS?
> 
> Some alternatives include:
> 
> * `overflow-x: scroll`, which is not the nicest, but works relatively well with touch pads at least.
> * [`min-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/min-content) sizing or similar, which would stretch the notification, potentially too much

Hi @agilgur5 
I did some changes to use recommended css override way to fix this issue, could you please help to review?

JessieTeng89:
@crenshaw-dev Kindly help to review this PR

JessieTeng89:
Hi @crenshaw-dev 
Can I know how to get this PR approved or if any other actions needed before merging it?
Thanks

JessieTeng89:
Hi @agilgur5
A gentle follow up with this PR, could you please help to review it?

agilgur5:
Yes, I had been planning to; I'm not sure why you asked several other uninvolved people [for review](https://github.com/argoproj/argo-ui/pull/546#issuecomment-2068031295) a day after you [made changes](https://github.com/argoproj/argo-ui/pull/546#discussion_r1573288510), which was a full 2 weeks after [changes were requested](https://github.com/argoproj/argo-ui/pull/546#discussion_r1553002577).

Please be patient, especially if you yourself do not respond immediately -- it would not be surprising if a maintainer took as long as you did to respond if not longer. And please interact with maintainers who already reviewed your PR, instead of asking others, which is counter-productive.

src/components/notifications/notifications.scss
```diff
@@ -0,0 +1,10 @@
+.Toastify__toast {
+    overflow-wrap: break-word;
+}
+
+.Toastify__toast-body {
+    min-width: 0;
+    div {
+        min-width: 0;
+    }
+}
```
agilgur5, line 10: I would not recommend overriding the component's styles. There are simpler ways of achieving this. This is also a global style, so it affects all usage of `Toastify`, not just the ones in the `Notifications` component. It should be scoped to the component if you were to do something like this.

src/components/notifications/notifications.tsx
agilgur5: I'd prefer to use `word-break: break-word;` than overriding the component's styles, it is less complicated.

But a simpler way to achieve this is to just add `max-width: 240px` in addition to the `overflow-wrap: 'break-word';`. You can put them on the `div` itself, instead of passing to the `toastMethod`.

I did confirm this works:
Screenshot 2024-06-10 at 11 06 08 PM

Since the toast's width is never changed, we can reliably set a fixed value for `max-width`. (although increasing the width of the toast might be a good idea for this problem)

agilgur5 (changes requested):
There are simpler ways of achieving this without overriding the component's styles, which can cause more problems and is not very reliable

src/components/notifications/notifications.tsx
```diff
@@ -21,6 +21,8 @@ export interface NotificationsProps {
     notifications: Observable<NotificationInfo>;
 }
 
+require('./notifications.scss');
```
agilgur5, line 24: I can't push to your branch and GH can't comment on unchanged lines, so here's what I was suggesting.

Remove this CSS file and import, instead, on line 50/52 below of the `div`, add:
```tsx
               <div onClick={(e) => {
                 // ... elided ... don't change anything with the `onClick`
                }} style={{
                    // fit long words
                    overflowWrap: 'break-word',
                    maxWidth: '240px'
                }}>
```

JessieTeng89: I totally agree, Thanks a lot, I will make the changes and let you know

src/components/notifications/notifications.tsx
```diff
@@ -47,6 +47,9 @@ export class Notifications extends React.Component<NotificationsProps> {
                         sel.removeAllRanges();
                         sel.addRange(range);
                     }
+                }} style={{
+                    overflowWrap: 'break-word',
```
agilgur5, line 51:
```suggestion
                    // fit long words by wrapping them instead of overflowing past the width
                    overflowWrap: 'break-word',
```

add the comment I mentioned [before](https://github.com/argoproj/argo-ui/pull/546#discussion_r1634122451)

agilgur5: I unfortunately can't commit that myself via suggestion as a maintainer normally can, I'm guessing because of Fidelity's permissions

JessieTeng89: Thanks, I will add the comment

JessieTeng89: Hi @agilgur5 
Just added explanatory comment as suggested

agilgur5 (approved):
LGTM. Thanks for noting and fixing this issue

gillesbouvier-qz:
Did this change add the following CSS?
```CSS
.Toastify__toast-body > div:last-child {
  word-break: break-word;
  ...
}
```

If so, it is not working as expected as it is breaking short words too. Did you mean to use this instead?
```CSS
    overflow-wrap: break-word;
```

agilgur5:
No, you can see the diff here, which is all of 4 LoC of an in-line style that uses `overflow-wrap: break-word`

PR #552, opened 2024-05-05
fix: Use type export for interface reexports

instauro:
Resolves the argo-ui related warnings described in https://github.com/argoproj/argo-workflows/issues/13009

Related webpack issue https://github.com/webpack/webpack/issues/7378 giving some background


agilgur5:
Ah right, I remember why I couldn't do this before, b/c it broke lint as I mentioned in https://github.com/argoproj/argo-workflows/pull/12514, since this repo was still on `tslint` which didn't support the `import type` syntax. But I did eventually unblock that with #509

agilgur5 (approved):
CI/lint+build pass now after #509. Thanks for the update!

Confirmed that [CD](https://github.com/argoproj/argo-cd/blob/653fc8da1f9c48cef535ff15e6ac7b5f795c9dae/ui/package.json#L121) and [Rollouts](https://github.com/argoproj/argo-rollouts/blob/125fc3dfbaec922209e5005fd4721d663042d1f3/ui/package.json#L25) are also on versions of TS that support this syntax

Issue #553, opened 2024-05-11
Replace custom components with `antd` components


agilgur5:
### Motivation

Reduce the scope of this component library, reduce its bundle size, use more tooling from the UI libraries we import, and have more consistent styling with that UI library. Should help with #453

We already use `antd`, let's use its components everywhere instead of having custom components in some places. At the very least, if we need custom functionality, we can wrap the `antd` components.

Also would be good to remove `foundation-sites` since that's another (dated) UI library and standardize on `antd`

### Use Cases

A few components that can be replaced from a quick glance:
1. [AutoComplete](https://ant.design/components/auto-complete)
1. [Checkbox](https://ant.design/components/checkbox)
1. [Dropdown](https://ant.design/components/dropdown)
1. [Form](https://ant.design/components/form)
1. [Notification](https://ant.design/components/notification)
1. `Popup` -> [Popover](https://ant.design/components/popover) / [Popconfirm](https://ant.design/components/popconfirm) / Modal / Alert
1. [Select](https://ant.design/components/select)
1. `SlidingPanel` -> [Drawer](https://ant.design/components/drawer)
1. [Tabs](https://ant.design/components/tabs)
1. [Tooltip](https://ant.design/components/tooltip)

Probably several others too

agilgur5:
> We already use `antd`

Oh this might not be correct actually. It's [only used in `./antd`](https://github.com/search?q=repo%3Aargoproj%2Fargo-ui%20antd&type=code) and nowhere else. That might explain how #387 managed to work.

Huh, I guess Workflows might not use `antd` at all, just installs it with `argo-ui` but does not end up bundling it

agilgur5:
Indeed there aren't even any components in `./antd`, there are only Storybook stories 😕

Apparently it was supposed to be used in v2: https://github.com/argoproj/argo-ui/pull/161#pullrequestreview-820560829, but that never happened (been 2.5 years since then now 😕)

Given that it's all unused code and unused deps, and complicates the build unnecessarily (another Storybook build, Less and Gulp usage for some reason, etc), and given the state of this repo per #453, I'm actually pretty inclined to remove it entirely.

What we may want to do in the future, per #453, is to switch all Argoproj UIs to use `antd` or another component library (`foundation-sites` is more a CSS library; there might be a wrapper component library but it's dated in any case)

agilgur5:
Closing as unplanned due to deprecation

Issue #453, opened 2023-09-25
`argo-ui`: future state of this repo


agilgur5:
This is based on discussions in [this Slack thread](https://cloud-native.slack.com/archives/C020XM04CUW/p1692085374857769) (which is fairly long and winding) that I had raised after realizing the fairly neglected and incomplete state of this repo when having to do an upgrade of it for Argo Workflows in https://github.com/argoproj/argo-workflows/pull/11585. 

This was also mentioned in a [Maintainers meeting](https://docs.google.com/document/d/1C71GA0XCxsyGP7Q4SYveHak5ZZcJOuaOiDCt_xHG10U/edit#heading=h.6upblgdtbf14) that I was not part of.

## Future State -- intent to phase out(?)

This repo has been neglected for some time and right now there are not that many cross-project contributors. Due to this and the interim state of this repo as not entirely a library (see below), this has resulted in slowing down development while simultaneously not really creating a "consistent look and feel" between projects (such as Workflows and CD). There are some similarities and shared branding, but many differences (in part because CD has received more attention) despite having some shared components.

While it may be possible to improve this situation, some contributors believe that this may not be worthwhile and that a design guide of sorts may be a better substitute. Argoproj also already uses [`antd`](https://github.com/ant-design/ant-design) as the underlying component library. This repo has higher-level components on top of those as well as some components that are not part of `antd`.
(Personally, I do not really have a strong opinion myself as I am a more recent contributor to Argoproj and am more documenting the current thoughts around this for everyone. I think the "ideal state" of this repo could be great, but if it does not suffice for those purposes and may not ever suffice, while also currently slowing things down, phasing out may certainly make a lot of sense.)

### Requiring downstream issues
In any case, as the repo is indeed not really actively maintained (and has no active maintainer list), we should document its current state.

Instead of having contributors create issues and PRs here ad-hoc, it may be better in the interim to require that issues and PRs have a downstream issue in consuming projects like Argo Workflows or Argo CD. That way, maintainers of the consuming projects will be aware of and can review PRs here.

### TypeScript compatibility issues
[...]

PR #161, opened 2021-10-21
Antd redesign

plakyda-codefresh:

PR #554, opened 2024-05-11
chore(deps): remove unused `antd` directory and deps

agilgur5:
Per https://github.com/argoproj/argo-ui/issues/553#issuecomment-2106048778 and https://github.com/argoproj/argo-ui/issues/453#issuecomment-2106051851, remove the unused `antd` dir and deps
- `antd` was seemingly meant to be used in `v2` per https://github.com/argoproj/argo-ui/pull/161#pullrequestreview-820560829, but that never happened.
  - It's been ~2.5 years since and per #453 this repo is effectively deprecated & unmaintained. Moving forward, Argoproj UIs should use `antd` directly instead of `argo-ui`, as Rollouts is doing in some places

### Motivation

- there are no components, only stories in the `./antd` dir
  - and doubly confirmed that none of these are used in CD, Workflows, or Rollouts

### Modifications

  - so remove the `./antd` dir entirely as well as all related deps
    - `antd` ofc, but also all the `less` deps and the `gulp` build for `antd`'s styles
      - I'm not even sure what the `gulp` build was for exactly, nor why it used different tooling than already existed in this repo
      
### Verification

The removed files etc are entirely unused


agilgur5:
Fixed merge conflict with #534

isubasinghe (approved):
Builds locally, given this is removal I am going to approve despite not being able to run storybook locally.

crenshaw-dev (approved):

agilgur5:
Fixed a tiny merge conflict in the `package.json#scripts` from #536

PR #556, opened 2024-05-13
feat(popup): styles for list element in popup

agaudreault:
Necessary for https://github.com/argoproj/argo-cd/pull/18116

image


agilgur5 (approved):
LGTM off hand, tiny refactor and tiny addition.

It sounds like we might want to apply some of these styles on _all_ children, instead of specific ones, but that can be a future improvement.

I'll let CD folks double-check.

ishitasequeira (approved):

PR #558, opened 2024-05-29
fix: copy logs text on Linux & Windows. Fixes #557

### LINKED ISSUE #557: Copy from workflow logs using Ctrl-C doesn't work
All the details for this issue are here https://github.com/argoproj/argo-workflows/issues/12947

I dug into this issue ans it's due to the `Terminal` from `xterm` in `LogsViewer` component that catches Keyboard events. [This bug ](https://github.com/xtermjs/xterm.js/issues/2478)is still in an open state on xterm repo so no real need to update it.
As a workaround I'll implement something inspired from https://github.com/xtermjs/xterm.js/issues/2478#issuecomment-1002371588

Adrien-D:
Fixes #557

Inspired from https://github.com/xtermjs/xterm.js/issues/2478#issuecomment-1002371588 this `attachCustomKeyEventHandler` on the terminal catch the `Ctrl+C` event, and if there is a current text selection, we force the browser to copy that text.

### DIFF
--- a/src/components/logs-viewer/logs-viewer.tsx
+++ b/src/components/logs-viewer/logs-viewer.tsx
@@ -43,6 +43,16 @@ export class LogsViewer extends React.Component<LogsViewerProps> {
         this.terminal.loadAddon(this.fitAddon);
         this.terminal.open(container);
         this.fitAddon.fit();
+        // handle Ctrl+C for copying logs
+        this.terminal.attachCustomKeyEventHandler((ev) => {
+            if (ev.ctrlKey && ev.code === "KeyC" && ev.type === "keydown") {
+                const selection = this.terminal.getSelection();
+                if (!selection) return true;
+                
+                navigator.clipboard?.writeText(selection);
+                return false
+            }
+        });
     }
 
     public componentDidMount() {


JPZ13:
@agilgur5 @terrytangyuan - who would be the best to ask for a review on this and other UI PRs? We have a couple UI PRs coming down the pipeline

terrytangyuan:
cc @argoproj/argo-ui-maintainers

agilgur5:
Followed up [on Slack](https://cloud-native.slack.com/archives/C0510EUH90V/p1717607518614589).

As per there, prior UI maintainers unfortunately do not respond and the GH team name doesn't work anymore either 😕 See also #453 et al.

I've been reviewing various code here since I've had Approver permissions in this repo, and have asked Crenshaw, Terry, and Remington for review before. Crenshaw has responded after I've pinged him or the formerly named `#argo-contributors` channel several times. Remington responded a bit in the past, but not recently. In recent times, I've asked Isitha for review too since we both have Approver permissions now and Isitha works a bit on UI in Workflows.

As I'm the only active UI SME on Workflows, it would be good if I reviewed more substantial changes. If I'm not available or it's a smaller change, I think others could respond.

I was also waiting for Adrien to respond to code review I'd done [last week](https://github.com/argoproj/argo-workflows/pull/13113#discussion_r1619379849) before adding more; he did so earlier today.

src/components/logs-viewer/logs-viewer.tsx
```diff
@@ -43,6 +43,18 @@ export class LogsViewer extends React.Component<LogsViewerProps> {
...
+                const selection = this.terminal.getSelection();
+                if (selection) {
+                    if (navigator.clipboard) {
+                        navigator.clipboard.writeText(selection);
+                    }
+                    return false;
+                }
+            }
+            return true;
+        });
```
agilgur5, line 57: let's invert this conditional for less nesting and less code complexity, similar to https://github.com/argoproj/argo-workflows/pull/11596, https://github.com/argoproj/argo-workflows/pull/12146, https://github.com/argoproj/argo-workflows/pull/13094, etc.

We can also simplify some pieces as this function returns `void`, such as with optional chaining.
Also `e` or `ev` is the common convention for event names.
And this could use a comment

```suggestion
        // handle Ctrl+C for copying logs
        this.terminal.attachCustomKeyEventHandler((ev) => {
            if (ev.ctrlKey && ev.code === "KeyC" && ev.type === "keydown") {
                const selection = this.terminal.getSelection();
                if (!selection) return;
                
                return navigator.clipboard?.writeText(selection);
            }
        });
```

agilgur5: Oh woops, the [`xterm` docs](https://xtermjs.org/docs/api/terminal/classes/terminal/#attachcustomkeyeventhandler) are a bit confusing as they actually didn't give _what_ the function should return

I double-checked against [the typings](https://github.com/xtermjs/xterm.js/blob/9ba6c00a195c95fcf8292a2b9084d91450e5daae/typings/xterm.d.ts#L1019) and apparently it does take a boolean, for whether the event should continue propagating or not to `xterm`. We don't really care about that part though as we're only using it as a log UI.

Adrien-D: Thanks for your review @agilgur5. Yes I did some nesting to handle the boolean returned value in case of selected text. But if we don't care about that as we are only logging then I'll implement your suggestion.

agilgur5: The boolean return is irrelevant for our use-case, but your version is more correct so sounds good.

Also I am pretty sure we can replace this component with something much smaller (per https://github.com/argoproj/argo-workflows/issues/12059, it's quite big) because we don't use even half of its features

agilgur5 (approved):
LGTM

PR #535, opened 2024-02-12
refactor(deps)!: remove `moment` dep and usage

agilgur5:
Related to the goals of https://github.com/argoproj/argo-workflows/issues/12059, just removing / replacing a large dep entirely.
Follow-up to https://github.com/argoproj/argo-workflows/pull/12097 and #534 here.
Needed for https://github.com/argoproj/argo-workflows/pull/12611

### Motivation

`moment` has been [deprecated since Sep 2020](https://momentjs.com/docs/#/-project-status/) and recommends using native `Intl` or newer libraries that make use of native `Intl`, such as `luxon` and `day.js`
- `moment` is also a very large dependency and hence is ripe for pruning and replacement as well

### Modifications

- remove `formatTimestamp` function from `v2/utils` as it is unused
  - it is unused [internally](https://github.com/search?q=repo%3Aargoproj%2Fargo-ui+formattimestamp&type=code) (it is only exported here), unused by [CD](https://github.com/search?q=repo%3Aargoproj%2Fargo-cd%20formattimestamp&type=code), unused by [Workflows](https://github.com/search?q=repo%3Aargoproj%2Fargo-workflows+formattimestamp&type=code)
    - it's also unused by Rollouts, which has [two](https://github.com/argoproj/argo-rollouts/blob/c688dd89a7f883c8c004b4024b390104b5be5770/ui/src/app/shared/utils/utils.tsx#L17) [internal](https://github.com/argoproj/argo-rollouts/blob/c688dd89a7f883c8c004b4024b390104b5be5770/ui/src/app/components/rollout/revision.tsx#L21) copies of this function

- change `Ticker` component to use plain `Date` instead of `moment`
  - downstream projects should get a type error on this which should make the switch straightforward

- change `Duration` component to use existing `formatDuration` function which doesn't use `moment`
  - this will look different than the previous version, but the formatting between the two should really be consistent in any case
  - other option was to just remove this component entirely, since this repo is not actively maintained
    - leaving the component in instead of deleting it gives a more gradual off-ramp
[...]

PR #560, opened 2024-06-04
fix: remove autoHighlight on autocomplete by default. Fixes #559

### LINKED ISSUE #559: Autocomplete component is too aggressive on opening
When opening the Autocomplete complete it auto highlight the first item, so when you hit enter it select the first item and not what you freely search into the input.

My proposal is to simply remove this behavior that is misleading users.

Adrien-D:
Fixes #559 

Simply add the `autoHighlight` to props to still be customisable


src/components/autocomplete/autocomplete.tsx
```diff
@@ -25,6 +25,7 @@ export interface AutocompleteProps {
     autoCompleteRef?: (api: AutocompleteApi) => any;
     filterSuggestions?: boolean;
     qeid?: string;
+    autoHighlight?: boolean;
```
agilgur5, line 28:
```suggestion
    autoHighlight?: true;
```

Let's keep the same behavior for backward compat

agilgur5 (changes requested):
I wouldn't suggest changing the default behavior as that is a breaking change. Unless you can confirm that all usage of this component does not intend to use `autoHighlight`, we should not assume that all prior usage was incorrect.

I'd leave the default as `true`, then, on components where `autoHighlight` is incorrect or not needed, we can explicitly set the new prop to `false`.

Adrien-D:
@agilgur5 can't really tell if all usage of this component are intend to use `autoHighlight`, but by passing the `props.autoHighlight` down to `ReactAutocomplete`, all the existing components will pass `undefined`, so it will take the `true` upstream value, so it shouldn't change the actual behavior and be breaking change. But if you consider that we can't trust this upstream value I can keep this on our side.

agilgur5:
> can't really tell if all usage of this component are intend to use `autoHighlight`

For reference, I've looked at all usage of components in Workflows, CD, and Rollouts when making changes (esp breaking ones), such as in #535 -- see my PR description there which lists all usage of it.
(technically there can be other consumers too, but it doesn't have stable releases so nbd about those)

> all the existing components will pass `undefined`, so it will take the `true` upstream value, so it shouldn't change the actual behavior and be breaking change.

Oh, you're correct. Then your PR title is not quite accurate and should be reworded; i.e. "fix(autocomplete): make `autoHighlight` configurable".


> But if you consider that we can't trust this upstream value I can keep this on our side.

Well [the `react-autocomplete` library](https://github.com/reactjs/react-autocomplete) hasn't had a commit in 6+ years and was archived 4 years ago, so I don't think it's changing any time soon 😅 

But having the default visible here helps with typings; it will show the default whereas without it, it's ambiguous to a consumer since `react-autocomplete`'s usage is encapsulated.

Technically speaking, we should use the type directly, i.e. with `ReactAutocomplete.Props['autoHighlight']`, except it turns out that [the typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1a61bba1934a71d481c2b56104799b9db5a4fcec/types/react-autocomplete/index.d.ts#L120) are missing [the default](https://github.com/reactjs/react-autocomplete/blob/41388f7d7760bf6cf38e7946e43d4fddd9c7c176/lib/Autocomplete.js#L188).

src/components/autocomplete/autocomplete.tsx
agilgur5:
```suggestion
    autoHighlight?: true; // ReactAutocomplete.Props['autoHighlight'] is missing the `true` default per https://github.com/argoproj/argo-ui/pull/560#issuecomment-2155872347
```

including my [below](https://github.com/argoproj/argo-ui/pull/560#issuecomment-2155872347) suggestion with a comment

agilgur5 (approved):

src/components/autocomplete/autocomplete.tsx
```diff
@@ -25,6 +25,7 @@ export interface AutocompleteProps {
     autoCompleteRef?: (api: AutocompleteApi) => any;
     filterSuggestions?: boolean;
     qeid?: string;
+    autoHighlight?: true; // ReactAutocomplete.Props['autoHighlight'] is missing the `true` default per https://github.com/argoproj/argo-ui/pull/560#issuecomment-2155872347
```
agilgur5, line 28: oh defaults may not be supported for TS interface per the tests in my PR to DT: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69761#discussion_r1631989188

it's been a while since I've done more nuanced typings, my bad if so

agilgur5: Yea

```tsx
<Autocomplete autoHighlight={false} />
```

has a typecheck error:
```
Type 'false' is not assignable to type 'true'.ts(2322
```

@Adrien-D would you mind PR'ing a fix to this by switching to JSDoc `@default`? I'd do it myself but then we'd need another reviewer with approval permissions 😅

src/components/autocomplete/autocomplete.tsx
agilgur5: Follow-up [thread], apparently interface defaults are unsupported, so will just use JSDoc `@default` annotation

src/components/autocomplete/autocomplete.tsx
Adrien-D: @agilgur5 Sure, [here is the fix ](https://github.com/argoproj/argo-ui/pull/561)

agilgur5: Thanks!

PR #561, opened 2024-06-10
fix(autocomplete): use JSDoc `@default` for `autoHighlight` prop

Adrien-D:
Follow-up to / Fixes https://github.com/argoproj/argo-ui/pull/560#discussion_r1631992541


agilgur5 (approved):
Thanks for helping fix my mistake!

PR #563, opened 2024-06-11
fix(select): right align arrow

tczhao:
https://github.com/argoproj/argo-ui/pull/178
had `position: absolute;` in all the test
but not in the commit 🤯

Fixes https://github.com/argoproj/argo-ui/pull/178

### DIFF
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -28,6 +28,7 @@
     
     &__value-arrow {
         // right most, vertically centered
+        position: absolute;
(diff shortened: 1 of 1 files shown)


agilgur5:
oh poop, that was my mistake, I apparently [accidentally removed it](https://github.com/argoproj/argo-ui/pull/178#discussion_r1634037052) when I added the comment. Thanks for catching that and fixing it!

agilgur5 (approved):

PR #565, opened 2024-06-12
fix(navigation): invalid redirection

### LINKED ISSUE #564: Navigation error with location.href ends with `/`
When I discover applications using `https://argocd.com/service/applications/`, search applications and then click it.
But it is redirected to `https://argocd.com/service/applications//{application-name}`. so the slash in the end has to be removed.

KSH-code:
fix #564

When I discover applications on like `https://argocd.com/service/applications/`, search applications and then click it.
But it is redirected to `https://argocd.com/service/applications//{application-name}`.
And an empty page is shown.

for the soultion, the slash in the end has to be removed.

### DIFF
--- a/src/components/navigation.ts
+++ b/src/components/navigation.ts
@@ -14,7 +14,9 @@ export class NavigationManager implements NavigationApi {
     }
 
     public goto(path: string, query: {[name: string]: any} = {}, options?: { event?: React.MouseEvent, replace?: boolean }): void {
-        if (path.startsWith('.')) {
+        if (path.startsWith('./') && this.history.location.pathname.endsWith('/')) {
+            path = this.history.location.pathname + path.slice(2);
+        } else if (path.startsWith('.')) {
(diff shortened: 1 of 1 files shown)


KSH-code:
Hi @agilgur5 
Could you please a feedback pr for merging?
Thanks!

agilgur5:
I can take a look, but I'm not a CD Maintainer so I wouldn't merge on my own in any case. It would be good to link to an issue upstream as well per #453

KSH-code:
Thank you for your check.
I've made an issue on https://github.com/argoproj/argo-cd/issues/19093 which include referring this.

agilgur5:
I left a comment on your upstream issue; this is perhaps better addressed upstream than here: https://github.com/argoproj/argo-cd/issues/19093#issuecomment-2240032528

KSH-code:
I agree. I close this so that I will address this issue on upstream.

PR #567, opened 2024-07-18
fix(deps-dev): downgrade all Storybook breakage from dependabot et al

### LINKED ISSUE #469: Project setup fails with module build error

agilgur5:
Fixes #469 

### Motivation

Dependabot made lots of breaking changes prior to #537 that made the Storybook unrunnable

#### devDeps
- go back to Storybook v6 addons, matching the `storybook` dep <details>
  - the addons were partially auto-upgraded to breaking versions by dependabot in #382, https://github.com/argoproj/argo-ui/pull/539#issuecomment-1944279905, https://github.com/argoproj/argo-ui/pull/541#issuecomment-1944282035, and others
  - why `v6.5.0-beta.1` specifically? See [below comment](https://github.com/argoproj/argo-ui/pull/567#issuecomment-2239925757)
  </details>
- go back to Webpack v4, matching Storybook v6
  - this was also auto-upgraded to a breaking version by dependabot: https://github.com/argoproj/argo-ui/pull/260#issuecomment-2235160404
  - go back to `ts-loader` v8, matching Webpack v4
    - and this too had a breaking dependabot upgrade: https://github.com/argoproj/argo-ui/pull/227#issuecomment-1879870942

- remove deprecated, no longer used `@types/storybook*` deps <details>
  - Storybook has types built-in now (and is written in TS, not sure if it wasn't before or something)
  </details>
  
- migrate from `@dump247/storybook-state` to `useState` hook in v1 <details>
  - as per its [archived README](https://github.com/dump247/storybook-state/tree/1da90ad?tab=readme-ov-file#storybook-state)
  - tried to make the diff as small as possible
  - in newer Storybook, this might make more sense to [use `Args`](https://storybook.js.org/docs/6/writing-stories/args) like v2 does. that's for future work though, as I don't believe `storiesOf` syntax even supports that
    </details>
    
#### others
- Per https://github.com/argoproj/argo-workflows/pull/11585, the `react-toastify` pin in #379 seemed to have been made to work with Webpack v5 in Argo CD, but this repo's Webpack was on v4 via Storybook v6 (per above)
  - Downstream consumers should force resolve `react-toastify` to `9.0.8` or other via `resolutions` etc if needed

- match `tsconfig.json` from #367

- import individual v1 components in stories instead of from `src/components`, speeding up `yarn start` from minutes to seconds by avoiding importing all other components due to side-effects (see https://github.com/argoproj/argo-workflows/pull/12158 for reference)

### Verification

Storybooks for v1 and v2 work now via `yarn start` and `yarn start-v2`:
<details><summary>Screenshots:</summary>

Screenshot 2024-07-18 at 3 51 31 PM
Screenshot 2024-07-18 at 3 54 13 PM

</details>
 
### Notes to Reviewers

~I could still not get `yarn start` to work. Got through a bunch of errors. I've put in too much time on attempting this, so I am giving up at this point since this repo is deprecated anyway per #453. Leaving this as a draft of my work for reference.~ **EDIT**: and apparently I was [one change](https://github.com/argoproj/argo-ui/pull/567#issuecomment-2235327831) from it working apparently 😅 

~I did also try upgrading everything to Storybook v7 or at least Storybook v6 with Webpack v5 in https://github.com/argoproj/argo-ui/pull/432#issuecomment-1879816619 and #568, and those had many more issues, so upgrading is not time well spent either.~ **EDIT**: now that this and #568 work, the upgrade might actually be doable. All the original dep mismatches etc might have been part of the issue.

The CSS does not work for v1's Storybook and I couldn't get it to work after [several attempts](https://github.com/argoproj/argo-ui/pull/567#issuecomment-2239556449), but this is still much better than the previous unrunnable state and v2 at least works now.

### Future Work

1. upgrade old Storybook v5 syntax and config -- see #568 
1. upgrade to Storybook v7 (and v8 now)
1. switch v1 stories' React `useState` to Storybook `Args`, per above
1. get CSS working for v1

### Postscript

It seems likely that the Storybook build has not been working for _years_, since some of the reverts are _years_ old and the incompatible changes made by humans would not have worked with the Storybook build. I am guessing it was only tested against Argo CD and the Storybook was unused.

If that is the case, it may be worth deleting the Storybook entirely as it's been broken and unused for years and not worth maintaining. And has not been used in the dev process for at least an equivalent amount of time


agilgur5:
> still erroring on one of Storybook's own imports

or, actually that looks like the error originates from `@dump247/storybook-state`, which hasn't been updated in 5 years, i.e. long deprecated, and which has long been giving a warning during install about an incorrect peerDep version on `@storybook/addons`:

```
warning " > @dump247/storybook-state@1.6.1" has incorrect peer dependency "@storybook/addons@^3.2.16".
```

and `@storybook/addons@3.2.16` is [6.5 years old](https://github.com/storybookjs/storybook/tree/v3.2.16)...

The upgrade path from it to `useState` is pretty easy, but I uh, have a feeling things won't start working after that either if it's been on an incorrect peer dep for _that_ long 😕

agilgur5:
> The upgrade path from it to `useState` is pretty easy, but I uh, have a feeling things won't start working after that either if it's been on an incorrect peer dep for _that_ long 😕

I commented some pieces out and, well, it did finally _build_, and then the Storybook window opened in the browser and completely froze.

Suffice to say, I am pretty partial to just nuking the Storybook entirely now, so there is no confusion about its state or usability and to continue reducing the scope of this codebase until it is no longer used.

agilgur5:
> it did finally _build_, and then the Storybook window opened in the browser and completely froze.

Huh it did actually load and work fine, after I waited 3 minutes of my fans spinning quite loudly. So this actually might work...
v2 started much quicker, so upgrading from `storiesOf` etc syntax in #568 might improve that poor performance

agilgur5:
> Huh it did actually load and work fine, after I waited 3 minutes of my fans spinning quite loudly.

despite this long load time, the CSS also did not load, which makes the v1 Storybook not very useful. the v2 one loads fine.
I tried a bunch of suggestions from Storybook issues and not one of them worked 😕
I saw that there were some in-line styles added in the component iframe, but they were all `<style>undefined</style>`, which definitely suggests some error is happening and not getting logged.
Also building a static Storybook (i.e. w/o dev server) had the exact same issues (no CSS but long load time).

This PR is still useful as-is though, so I guess we'll leave that for another PR or maybe the upgrade to Storybook v7 or v8 or Webpack v5 will resolve this.

> v2 started much quicker, so upgrading from `storiesOf` etc syntax in #568 might improve that poor performance

The biggest difference seemed to not be the upgrade from `storiesOf` syntax but actually individual component imports per https://github.com/argoproj/argo-workflows/pull/12158. Then it seemed that importing the global CSS was the main cause of it slowing down, despite the import doing nothing per above 🫠 .

agilgur5:
Huh, CI [has a failure](https://github.com/argoproj/argo-ui/actions/runs/10011139944/job/27673910285?pr=567) that I don't have locally:

```
ERR! TypeError: Cannot read properties of null (reading 'babel')
ERR!     at _callee$ (/home/runner/work/argo-ui/argo-ui/node_modules/@storybook/telemetry/dist/cjs/storybook-metadata.js:202:52)
ERR!     at tryCatch (/home/runner/work/argo-ui/argo-ui/node_modules/regenerator-runtime/runtime.js:64:40)
ERR!     at Generator.invoke (/home/runner/work/argo-ui/argo-ui/node_modules/regenerator-runtime/runtime.js:299:22)
ERR!     at Generator.next (/home/runner/work/argo-ui/argo-ui/node_modules/regenerator-runtime/runtime.js:124:21)
ERR!     at asyncGeneratorStep (/home/runner/work/argo-ui/argo-ui/node_modules/@storybook/telemetry/dist/cjs/storybook-metadata.js:68:103)
ERR!     at _next (/home/runner/work/argo-ui/argo-ui/node_modules/@storybook/telemetry/dist/cjs/storybook-metadata.js:70:194)
ERR!     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

agilgur5:
> Huh, CI [has a failure](https://github.com/argoproj/argo-ui/actions/runs/10011139944/job/27673910285?pr=567) that I don't have locally:

Ah it only happens on `build`, but not on `start`; that I can repro. Also same error in [this unanswered SO question](https://stackoverflow.com/q/76572293/3431180). The error traces back to [this line in Storybook code](https://github.com/storybookjs/storybook/blob/7db13fface8f05b870a16b90694a840785830f7e/lib/telemetry/src/storybook-metadata.ts#L113) where I'm guessing `mainConfig` refers to `main.js`, which does not exist in v1. Notably #568 does not have this error, and it moves to using `main.js`

agilgur5:
> The error traces back to [this line in Storybook code](https://github.com/storybookjs/storybook/blob/7db13fface8f05b870a16b90694a840785830f7e/lib/telemetry/src/storybook-metadata.ts#L113) where I'm guessing `mainConfig` refers to `main.js`, which does not exist in v1.

I tried downgrading to Storybook `6.4.22` (latest of 6.4.x) which does not have `@storybook/telemetry` (the line above was introduced in https://github.com/storybookjs/storybook/pull/18046), but then started getting a typecheck error on `react-router` for some reason 😕  Then upgraded to `6.5.0-alpha.64` and the Storybook failed to load with some errors in the browser console. Then upgraded to `6.5.0-beta.1` and had the same issue, but once I reset my `node_modules`, it went away (that's happened twice now; stale cache with Yarn or something?)

> Also same error in [this unanswered SO question](https://stackoverflow.com/q/76572293/3431180).

I [answered it myself](https://stackoverflow.com/a/78770854/3431180) 😅

agilgur5:
> I've put in too much time on attempting this, so I am giving up at this point since this repo is deprecated anyway per #453.

> Suffice to say, I am pretty partial to just deleting the Storybook entirely now, so there is no confusion about its state or usability and to continue reducing the scope of this codebase until it is no longer used.

For better or for worse, I had a pretty bad migraine all of yesterday (and a bit of the night before and this morning), so I worked on this yesterday as it was something fairly straightforward that didn't require a lot of brainpower, i.e. trial, wait, error, trial, wait, error, repeat until working with right versions, config, etc. Tedious af, but not really "difficult" per se. 🙃

ashutosh16:
thank you @agilgur5 for putting the effort. I agree with you comments and I think this is good starting point to get the storybook fix.   I'm not sure if the dependent bot will update the dependencies  and break again.

agilgur5:
> I'm not sure if the dependent bot will update the dependencies and break again.

It won't, it now only runs for security fixes: #537

ashutosh16 (approved):
LGTM

crenshaw-dev (approved):
Local test checks out. Thanks so much @agilgur5!

PR #569, opened 2024-07-20
chore(deps-dev): remove unused `@testing-library/` deps

agilgur5:
### Motivation

- they're not used in the repo -- can search the repo for `@testing-library` and get no results other than the dep manifests
  - I guess I missed these in my previous removals (#534) and assumed they were used or something
  - there aren't that many tests in this repo either
    - while `testing-library` is the preferred tool in the ecosystem these days, `enzyme` is still used in this repo
      - potentially could be replaced by `testing-library` (if no mounting etc is necessarily used), but not necessarily worthwhile given the deprecated state of this repo (#453)

### Modifications

- remove `@testing-library/` deps from the `package.json` and `yarn.lock` via `yarn remove`

- also automatically reorder the `@types/uuid` dep
  - happens automatically on a `yarn add` or `yarn remove`, I just left it out from previous removals as they were much larger and so it could be distracting
  
### Verification

Build, lint, test, etc all still function fine as CI shows


crenshaw-dev (changes requested):
I rebased locally and am getting errors from both `yarn install` and `yarn start`. Can you take a look?

agilgur5:
Rebased it myself here and it's working for me locally

crenshaw-dev (approved):
Local tests look good!

PR #571, opened 2024-07-20
chore(deps): dedupe `yarn.lock`, add check for dupes

agilgur5:
Equivalent to https://github.com/argoproj/argo-workflows/pull/11637 and PRs I've made in other repos

### Motivation

Same as downstream et al, see https://github.com/atlassian/yarn-deduplicate#why-is-this-necessary 
Removes ~2.5k+ LoC from the `yarn.lock` by consolidating duplicated deps (and saves on disk, installation, network, dep issues, etc)

In particular for this repo, all the dep removals and changes I've done for #569 #534 #567 etc have complicated `yarn.lock` diffs that I've manually reduced (sometimes a removal even causes an addition because `yarn` just adds the latest version of the dep). This does it automatically and will ensure it doesn't happen again.

I've wanted to do it for a while but it usually merge conflicts with any other dep changes so I kept deferring it and then forgetting it 😅 There should be a lot less of those after #537 though (dependabot also most certainly contributed to duped deps)

### Modifications

- add a `deduplicate` script that uses `yarn-deduplicate`
- run `deduplicate` during UI CI
- run `deduplicate` once and commit the results

- move the `build` script to be in the equivalent place as the `build-v2` script, after `start`/`start-v2`

### Verification

- `yarn deduplicate` does not cause another diff
- CI should pass
- (`yarn start` is still broken until #567 is merged)

### DIFF
--- a/.github/workflows/build-v1.yml
+++ b/.github/workflows/build-v1.yml
@@ -15,6 +15,7 @@ jobs:
         with:
           node-version-file: ".nvmrc"
       - run: yarn install
+      - run: yarn deduplicate
       - run: yarn build
       - run: yarn lint
       - run: yarn tsc -p tsconfig.json
--- a/package.json
+++ b/package.json
@@ -9,11 +9,12 @@
         "v2"
     ],
     "scripts": {
-        "build": "NODE_OPTIONS='--openssl-legacy-provider' build-storybook -o ./dist/storybook",
         "lint": "eslint --ext .tsx .",
+        "deduplicate": "npx yarn-deduplicate -s fewer yarn.lock",
         "test": "jest",
         "utils:icons": "rm -f src/assets/fonts/* && node ./scripts/icons/generator.js",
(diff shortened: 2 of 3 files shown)


agilgur5:
Ah a [build failure](https://github.com/argoproj/argo-ui/actions/runs/10022803913/job/27702989198?pr=571) in the deps, same as the ones I dealt with heavily manually before 🫠 

This one appears to be in the dep: https://github.com/iarna/wide-align/issues/63. Will resolve to a lower version instead of a higher one from the dedupe to get around it

crenshaw-dev:
Can you rebase?

agilgur5:
Rebased.

crenshaw-dev:
Sorry, gotta ask for another rebase 😆

agilgur5:
Rebased once more 😅

crenshaw-dev (approved):

PR #568, opened 2024-07-18
chore(storybook): update v1 syntax to use Storybook v6

agilgur5:
Follow-up to https://github.com/argoproj/argo-ui/pull/432#issuecomment-1879816619 and make the v1 code Storybook v7 compatible. This is just one step toward the upgrade, but perhaps the most breaking one.
Follow-up to #567 "Future Work" part 1.

### Motivation

- while the deps used Storybook v6, the configuration and stories themselves were still on a legacy v5 format
  - v6 had backward compat for v5, but v7 does not, so this needs upgrading/migrating
    - or well, v7 has some legacy mode for it that can be enabled with some config, but it is entirely gone in v8

### Modifications

 - migrate `storiesOf` to CSF per https://storybook.js.org/docs/7/migration-guide#storiesof-to-csf <details>
    - then had to do a bunch of manual changes to get back mostly the same previous indentation (with some minor differences where there were mistakes or inconsistencies)
    - CSF is still valid through to latest v8
    </details>
 - also remove `index.stories.tsx` and replace with naming other story files `*.stories.tsx`
    
- migrate v1 `config.js` + `webpack.config.js` to Storybook v6+ `main.js` + `preview.js`
  - remove `ts-loader` and instead use the built-in TS support, with a few optimizations per https://github.com/storybookjs/storybook/issues/22164#issuecomment-1603627308
  
### Verification

`yarn start` works the same as it did in #567
 
### Notes to Reviewers
 
~This requires #567 to work and so is built on top of it. Please **do not merge** until that is merged first!~
Rebased on top

PR #576, opened 2024-08-21
feat: add action menu extension

ashutosh16:
Ref: https://github.com/argoproj/argo-cd/pull/19620

Screenshot:
Screenshot 2024-08-20 at 22 24 00


leoluz (changes requested):
Please update the UI extensions documentation explaining the new type

rbreeze (approved):
LGTM once documentation is added

stories/top-bar.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { TopBar } from '../src/components/top-bar/top-bar';
+
+
+storiesOf('TopBar', module)
```
agilgur5, line 6: So I do have a pending PR to update all this old syntax: #568 . I'd prefer not to have more stories before that is merged

agilgur5 (commented):
What is this extension adding? In the screenshot is it the new button? That wouldn't need a change in `argo-ui`.

I would also strongly advise against CD UI extensions reaching into `argo-ui` given the deprecated state of this repo per pinned issue #453 . Giving a deprecated repo a new feature that user-facing extensions hook into sounds like exposing far too much, especially for a not very reliable repo.
 I would also advise against that as this repo is not supposed to be specific to CD.

src/components/top-bar/top-bar.tsx
```diff
@@ -85,22 +86,26 @@ const renderBreadcrumbs = (breadcrumbs: { title: string | React.ReactNode, path?
...
-const renderActionMenu = (actionMenu: ActionMenu) => (
+const renderActionMenu = (actionMenu: ActionMenu, actionMenuExtensions: React.ReactNode[]) => (
     <div>
         {actionMenu.items.map((item, i) => (
             <button disabled={!!item.disabled} qe-id={item.qeId} className='argo-button argo-button--base' onClick={() => item.action()} style={{marginRight: 2}} key={i}>
                 {item.iconClassName && (<i className={item.iconClassName} style={{marginLeft: '-5px', marginRight: '5px'}}/>)}
                 {item.title}
             </button>
         ))}
+        {actionMenuExtensions && actionMenuExtensions.map((ext, index) => (
```
agilgur5, line 97: Adding another button does not require an `argo-ui` change, you can just add another `item` to the list in CD.

agilgur5 (changes requested):

stories/top-bar.stories.tsx
ashutosh16: agree, i have remove the stories and will add when the pending  PRs get merged

src/components/top-bar/top-bar.tsx
ashutosh16: I still want to pursue these changes, and I think this is an easy solution that does not require any extensions or argo. If we have a better solution in the future, I will be happy to review it.

agilgur5: > I think this is an easy solution that does not require any extensions or argo

It's definitely not "easy" given how invasive it is, and it adds extensions to `argo-ui`, which does in fact add logic to Argo

agilgur5: See the concrete guidance I left downstream for how to avoid a change in `argo-ui`: https://github.com/argoproj/argo-cd/pull/19620/files/3e67e6dbda7c7f716a35164c636ad174df0e7396#r1727992070

agilgur5:
I think this can be closed given the latest changes in https://github.com/argoproj/argo-cd/pull/19620 that obviate the need for this per [my suggestions](https://github.com/argoproj/argo-ui/pull/576#discussion_r1727997681).

CD folks should probably double-check my review there since I'm not a regular contributor there (nor have approver permissions there)

PR #588, opened 2025-03-27
fix: prevent double slashes in path handling - NavigationManager.goto

armanossiloko:
This PR attempts fix duplicate slashes in path handling in the `NavigationManager.goto` method.

I encountered a case where the `Search applications...` form in ArgoCD would display some suggested apps - if I click on an app, it leads me to `www.example.com/applications//my-apps` which leads to nothing being loaded on the page except the sidebar.

- Adjusted the `NavigationManager.goto` method to handle redirection to `./` paths when the current URL ends with a slash - to avoid ending up with duplicate slashes in the URL

Signed-off-by: Arman Ossi Loko armanossiloko@gmail.com


armanossiloko:
The email has to be wrapped in a `<>`, are you bloody serious?

PR #589, opened 2025-03-27
fix: prevent double slashes in path handling - NavigationManager.goto

armanossiloko:
This PR attempts fix duplicate slashes in path handling in the `NavigationManager.goto` method.

I encountered a case where the `Search applications...` form in ArgoCD would display some suggested apps - if I click on an app, it leads me to `www.example.com/applications//my-apps` which leads to nothing being loaded on the page except the sidebar.

- Adjusted the `NavigationManager.goto` method to handle redirection to `./` paths when the current URL ends with a slash - to avoid ending up with duplicate slashes in the URL

Signed-off-by: Arman Ossi Loko <armanossiloko@gmail.com>


src/components/navigation.ts
```diff
@@ -15,7 +15,11 @@ export class NavigationManager implements NavigationApi {
 
     public goto(path: string, query: {[name: string]: any} = {}, options?: { event?: React.MouseEvent, replace?: boolean }): void {
         if (path.startsWith('.')) {
-            path = this.history.location.pathname + path.slice(1);
+            if (this.history.location.pathname.endsWith('/') && path.startsWith('./')) {
+                path = this.history.location.pathname + path.slice(2);
+            } else {
+                path = this.history.location.pathname + path.slice(1);
+            }
```
jsoref, line 22: Here's what I used for testing this logic to try to understand things:
```js
> z='./foo/'
'./foo/'
> z.endsWith('/')
true
> z.startsWith('./')
true
> z.slice(2)
'foo/'
```

If this is what we're seeing, then, it makes enough sense to me.

Certainly this problem hits me a lot.

jsoref (approved):

src/components/navigation.ts
armanossiloko: Looks good enough!

PR #614, opened 2026-04-01
chore!: React 19 upgrade

jwinters01:
# Upgrades Repo to React-19

Part 1 to closing argoproj/argo-cd#25203

## Notable changes
There's a lot of changes in this PR, a lot of them are just minor API changes with the React 19 upgrade, but some key big moves are outlined below.

### react-form Compatibility Layer (src/components/form/compat.tsx) — New file

  The react-form package (which used class-based internals incompatible with React 19) was removed and replaced with a custom hooks-based implementation that preserves the same public API surface.

  Implements the following:
  - <Form> component with defaultValues, validateError, onSubmit, and getApi callbacks — matching the old react-form API
  - FormApi object with values, errors, touched, setValue, setError, submitForm, getFormState, setAllValues, setFormState
  - FormField HOC (withFieldApi) that connects any component to the form via React context
  - Built-in field components: Text, TextArea, Checkbox
  - Supports both context-based and prop-based (formApi prop) form API injection
  - Uses Proxy-like getter/setter pattern on the FormApi object so that mutations to api.values, api.errors, api.touched trigger re-renders (important for callers that mutate rather than call setters)
  - Exported both as ReactForm.* namespace and as named exports for drop-in compatibility

### Autocomplete Re-implementation (src/components/autocomplete/autocomplete.tsx)

  Replaced the deprecated react-autocomplete library with a custom implementation built on downshift's useCombobox hook:
  - Full re-write from class-based ReactAutocomplete to a functional hooks-based component
  - Menu positioning logic (including flip-to-top when near viewport edge) is now owned directly in the component rather than monkey-patched onto the old library's class instance
  - Listens to both scroll and resize events for dynamic repositioning
  - Preserves the public AutocompleteProps API (filterSuggestions, autoHighlight, renderItem, renderInput, autoCompleteRef, onSelect, onChange, etc.)

### Exports (src/components/index.ts)

  All compatibility layer types and components are re-exported from the package's main entry point, allowing argo-cd to import from argo-ui without changes.

### Enzyme Sunset

  Enzyme is no longer supported for React 19, it needed to be removed.

### Other Changes

  - Package dependencies updated (React 19, Storybook upgrades, downshift added, react-autocomplete/react-form removed)
  - Test files migrated from .spec.tsx → .test.tsx convention
  - Minor React 19 compatibility fixes across various components (ref handling, prop changes)


pjiang-dev:
@linghaoSu @choejwoo @keithchong 
If you folks have some time we would appreciate help on reviews for the 2 PRs for react 19 upgrade. Thanks!

pjiang-dev (commented):
Since other argo project like Workflows, Rollouts also use argo-ui, what happens when we merge this PR?
Will those projects still be stable or will they also need follow up PRs to upgrade to react 19?

src/components/form/compat.tsx
```diff
@@ -0,0 +1,297 @@
...
+// NestedForm is a legacy wrapper that scoped fields under a prefix key.
+// In practice it is always wrapping a standalone <Form> with its own state,
+// so rendering children directly is sufficient.
+export function NestedForm({children}: {field: string; children: React.ReactNode}) {
+    return <>{children}</>;
+}
+
+export function Form(props: FormProps) {
+    const [values, setValues] = React.useState<FormValues>(props.defaultValues || {});
+    const [touched, setTouched] = React.useState<Record<string, boolean>>({});
```
pjiang-dev: Maybe a bug here:
touched is stored as a flat Record<string, boolean> keyed by the dotted path (e.g. {'spec.source.repoURL': true}), but form-field.tsx:46 reads it with getNestedField(props.formApi.touched, props.field) which walks touched.spec.source.repoURL and gets undefined. So the touched && error gate is always false and inline error messages never render for any nested field. 
The old react-form v2 stored touched as a nested object tree, which is why this code worked before.

Fix: store touched as a nested tree using the same deepSet helper that's already used for values.

Also lets add a test in compat.test.tsx that submits a form with a spec.source.repoURL-style field and asserts the inline error appears after blur

src/components/autocomplete/autocomplete.tsx
```diff
@@ -40,92 +39,157 @@ export const Autocomplete = (props: AutocompleteProps) => {
...
         document.addEventListener('scroll', listener, true);
+        window.addEventListener('resize', listener);
         return () => {
-            document.removeEventListener('scroll', listener);
+            document.removeEventListener('scroll', listener, true);
+            window.removeEventListener('resize', listener);
         };
+    }, [isOpen, setMenuPositions]);
+
+    useEffect(() => {
```
pjiang-dev, line 129: props is a fresh object reference every render, so the effect re-runs every render. Every render the parent receives a new `{refresh: ...}` API object via its callback. 

argo-cd's repos-list and project-details pass autoCompleteRef and store the API. If they stored the API in state, it could cause an infinite render loop. Worth grepping to confirm no consumer is doing that.

src/components/form/compat.tsx
choejwoo: +1, and I think `submitForm` needs to also mark fields as touched. Tested both flat and nested fields.
- Flat field after blur: partially works
- Flat field with no blur, click Submit: partially works (e.g. Login form with empty `Sign In` click)
- Nested field after blur: broken (the bug you described)

jwinters01:
> Since other argo project like Workflows, Rollouts also use argo-ui, what happens when we merge this PR? Will those projects still be stable or will they also need follow up PRs to upgrade to react 19?

consumers of argo-ui typically consume it by pulling in a pinned commit hash ([workflows](https://github.com/argoproj/argo-workflows/blob/90a3252f386c9bcbf9fee9d53aede8a3cfa8f9fa/ui/package.json#L19) and [rollouts](https://github.com/argoproj/argo-rollouts/blob/master/ui/package.json#L12), for example), so they would be unaffected by these changes as it stands. That said, it is true that if they wanted to update their argo-ui import they would need to upgrade as well. We had decided in a maintainers meeting that this repo has pretty low volume in terms of changes, so the urgency for these repos to need to upgrade to accomodate this is pretty low.

pjiang-dev (approved):

src/components/form/compat.tsx
jwinters01: @choejwoo can you give me an example where nested fields are broken? I want to validate a fix. @pjiang-dev I addressed your comments here

src/components/autocomplete/autocomplete.tsx
jwinters01: instead of grepping to make sure a consumer isn't problematic I made the component itself more resilient to this kind of issue 👍

src/components/form/compat.tsx
choejwoo: @jwinters01 
I was testing on the `NEW APP` form (Applications list → NEW APP). Expected the `required` warning to show in two cases.
- After blurring the empty Application Name field
- After clicking Create on an empty form

Neither showed any warning. Haven't retested after your latest commit yet.


#### expected behavior
https://github.com/user-attachments/assets/b9d46ed9-7347-48fb-8818-7a30328dab11

#### broken case
https://github.com/user-attachments/assets/dc3c46eb-f110-4c78-a134-9ec873d7931e

choejwoo: After testing with the latest commit, the empty-form Create case looks fixed now. It seems the blur scenario still doesn't work.

https://github.com/user-attachments/assets/3b1ae141-ffc3-46ef-8cea-3e0be7658866

package.json
```diff
@@ -12,25 +12,23 @@
...
-        "start-v2": "NODE_OPTIONS='--openssl-legacy-provider' start-storybook -c ./v2/.storybook -s ./v2/.storybook/images -p 6006",
-        "build-v2": "NODE_OPTIONS='--openssl-legacy-provider' build-storybook -o ./dist/storybook-v2 -c ./v2/.storybook"
+        "start": "storybook dev -p 6006",
+        "build": "storybook build -o ./dist/storybook",
+        "start-v2": "storybook dev -c ./v2/.storybook -p 6006",
+        "build-v2": "storybook build -o ./dist/storybook-v2 -c ./v2/.storybook"
     },
     "dependencies": {
         "@fortawesome/fontawesome-free": "^6.2.1",
         "@tippy.js/react": "^3.1.1",
```
choejwoo, line 24:
```
Accessing element.ref was removed in React 19. ref is now a regular prop. 
It will be removed from the JSX Element type in a future release.
```
Not a React 19 blocker, but flagging for awareness. Tooltip library `@tippy.js/react` still uses the deprecated `element.ref` pattern (warning above). Upstream is archived, so we'll likely need to migrate to another library like [Floating UI](https://github.com/atomiks/tippyjs-react/commit/2699f0450c28a92cd5bbd402573ce9ed64252899) in a future React bump.

src/components/page/page.tsx
```diff
@@ -39,11 +44,11 @@ export const Page = (props: PageProps) => {
...
-                                return (
-                                    <Helmet>
-                                        <title>{titleParts.join(' - ')}</title>
-                                    </Helmet>
-                                );
+                                const nextTitle = titleParts.join(' - ');
+                                if (documentTitle !== nextTitle) {
+                                    setDocumentTitle(nextTitle);
+                                }
+                                return null;
```
choejwoo: I understand `react-helmet` was likely removed because it's incompatible with React 19. 
However, calling `setDocumentTitle` inside the `Context.Consumer` render callback triggers a new React 19 warning.
```
page.tsx:59 Cannot update a component (`Page`) 
while rendering a different component (`Context.Consumer`). 
To locate the bad setState() call inside `Context.Consumer`, 
follow the stack trace as described in https://react.dev/link/setstate-in-render
```
`documentTitle !== nextTitle` guard enough to mitigate this? or should we move the title update into a `useEffect` to keep the render fully pure?

src/components/form/compat.tsx
jwinters01: thanks Ill take a look at these 👍

package.json
jwinters01: I have started a list of non-blocking improvements to make once these PR's are merged, once they are I'll make some followup PR's (or create followup issues if the changes are nontrivial) to resolve them

src/components/page/page.tsx
jwinters01: I put in a fix for this. Just ended up pulling out all the extraneous things that complicated the system.

src/components/form/compat.tsx
jwinters01: I was able to validate a fix for this:

https://github.com/user-attachments/assets/bc606185-92f4-4893-8877-69e9f9c6f7eb

CHANGE_ANALYSIS.md
agaudreault, line 1: Remove.

agaudreault (approved):

choejwoo:
@jwinters01, Thanks for the updates! 😊

PR #618, opened 2026-05-14
fix: guard keypress handler against unregistered keyCodes

rickbrouwer:
This PR adds a guard in the keypress handler in `v2/shared/keypress.tsx` against accessing an unregistered key entry.

Root cause: when `addKeybinding` is called with an array of keys and `combo` falsy, it sets `groupForKey[key] = index` for every key in the array, but `groups[index][key]` is only populated for the last key (intermediate writes to `groups[index]` get overwritten on each iteration). As a result, pressing any non-last key from such an array makes `groupForKey[e.keyCode]` resolve to a group that exists but does not contain that specific key, so `groups[g][e.keyCode].pressed = true` throws.

This shows up via `argo-cd` registering `Object.values(NumKey)`and `Object.values(NumPadKey)` in the applications tiles and applicationsets list views. By pressing for example `*` (browser `keyCode 56`, i.e. the `8` key) while one of those views is mounted produces the screenshots below.

The crash is visible during development via webpack-dev-server's error overlay. In a production build it is swallowed silently into the console, which is why this has gone unnoticed.

Screenshots web development:
image

and my production server:

Image

### DIFF
--- a/v2/shared/keypress.tsx
+++ b/v2/shared/keypress.tsx
@@ -116,7 +116,7 @@ const handlePress = (e: KeyboardEvent, state: GroupMap) => {
     const {groups, groupForKey} = state;
     const g = groupForKey[e.keyCode];
 
-    if (groups[g]) {
+    if (groups[g] && groups[g][e.keyCode]) {
         let allPressed = true;
         groups[g][e.keyCode].pressed = true;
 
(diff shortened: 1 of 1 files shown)


choejwoo (approved):

blakepettersson (approved):

blakepettersson:
This has already been fixed with #624 🙈

PR #621, opened 2026-05-27
fix(form): mirror state writes to refs synchronously in compat layer

jwinters01:
Each setter (setValue, setAllValues, setError, setTouched, setFormState, resetAll, and the proxy setters on values/touched/errors) now updates valuesRef.current / touchedRef.current / errorsRef.current synchronously in addition to scheduling the React state update. This fixes an issue where, if a consumer reads the form state in the same handler as they set state, they will end up reading stale values.


choejwoo (approved):
LGTM!

crenshaw-dev (approved):

PR #624, opened 2026-06-24
chore: fix Rules-of-React violations for React Compiler readiness

jwinters01:
This PR wires `eslint-plugin-react-hooks` v7 (`recommended-latest`) into the
ESLint config. Beyond the classic exhaustive-deps check, v7 ships the Rules-of-
React lints the React Compiler depends on: it flags refs read during render,
`setState` called synchronously in effects, mutation of values React treats as
immutable, and conditionally-called hooks.

This is the prerequisite for the React Compiler. The compiler only memoizes a
component when it can prove the component obeys the Rules of React; where it
can't, it silently bails on that component and the optimization is lost. Since
argo-ui is consumed as source and compiled by argo-cd's build, the compiler runs
on the consumer side — the library's job is to be violation-free so the
consumer's compiler doesn't bail on these components. Each violation below is one
such component. The consumer-side compiler integration lives in
argoproj/argo-cd#28417. This PR fixes only the error-level findings; the exhaustive-deps
warnings are left for follow-up.

### Classes of changes

**`rules-of-hooks` — hooks must run unconditionally on every render.**
`props.theme || useTheme()` short-circuited, so `useTheme()` was called
conditionally. Split into an unconditional call, then the fallback.
[theme-div.tsx](https://github.com/jwinters01/argo-ui/blob/781fa5e2ea74be92e1d5260e0119805c888858e4/v2/components/theme-div/theme-div.tsx#L21)

**`immutability` — don't reassign values React owns.** A render-scoped `let` was
reassigned from a function that returned a fresh object; the function now mutates
and returns the same object and callers use `const`.
[keypress.tsx](https://github.com/jwinters01/argo-ui/blob/781fa5e2ea74be92e1d5260e0119805c888858e4/v2/shared/keypress.tsx#L247)

**`refs` — don't read `ref.current` (or the ref prop) during render.**
`ref={props.innerref}` read the ref off props in JSX; now `innerref` is
destructured and the rest spread.
[input.tsx](https://github.com/jwinters01/argo-ui/blob/781fa5e2ea74be92e1d5260e0119805c888858e4/v2/components/input/input.tsx#L55)

**`set-state-in-effect` — derive state instead of an effect + `setState`.** An
effect synced the controlled `value` from `props.value`; replaced with the
derive-during-render `prevPropValue` pattern, leaving the effect to fire only
`onChange`.
[checkbox.tsx](https://github.com/jwinters01/argo-ui/blob/781fa5e2ea74be92e1d5260e0119805c888858e4/v2/components/checkbox/checkbox.tsx#L12)


agaudreault (approved):

jwinters01:
@choejwoo @blakepettersson would you all be able to take a look at this PR and the argo-cd one? It is the groundwork for integrating the react compiler into argo-cd

blakepettersson:
I'll approve if you can take a look at argoproj/argo-cd#28005 @jwinters01 😉

choejwoo (approved):
I verified some changes in Storybook and they work as expected. LGTM!

blakepettersson (approved):
Looks great!

PR #626, opened 2026-07-22
fix: update react-toastify for React 19 compatibility

choejwoo:
Related to argoproj/argo-cd#28749

Update `react-toastify` from v9.0.3 to v11.1.0 for React 19 compatibility.

React Toastify v9 relies on function component `defaultProps`, which are not applied when using React 19 with the automatic JSX runtime. This causes `transition` to be `undefined` and crashes the UI whenever a toast is rendered.

This also replaces the removed `toast.POSITION.BOTTOM_RIGHT` constant with the equivalent `bottom-right` position value.

After this PR is merged, I will update the `argo-ui` commit referenced by Argo CD and align its direct `react-toastify` dependency with v11 in a follow-up PR.

스크린샷 2026-07-22 오후 7 46 34


blakepettersson (approved):
LGTM

PR #632, opened 2026-09-01
fix: restore maxWidth prop for Tooltip

MasonM:
The changes in https://github.com/argoproj/argo-ui/pull/627 meant `<Tooltip>` no longer accepts the `maxWidth` prop, which Argo Workflows uses: https://github.com/argoproj/argo-workflows/blob/2434a0c5d4dcc20fd2c8c8407682e47105aede9f/ui/src/shared/components/tooltip.tsx#L38

(cc @jwinters01)


src/components/tooltip/tooltip.tsx
choejwoo, line 132: Should `maxWidth` be added here and to the dep array too? Otherwise it only applies on the initial `tippy()` call.

choejwoo (commented):
I don't know Argo Workflows well, but I noticed `boundary='viewport'` on that same line too. Is that one fine as is?

src/components/tooltip/tooltip.tsx
MasonM: Yep, good find. Done!

Joibel (approved):

choejwoo (approved):

PR #635, opened 2026-09-13
feat: add skipIfUnchanged option to NavigationManager.goto

rickbrouwer:
Filters in the Argo CD UI report their state up when they mount, so a remount writes the same query string once per filter. Browsers rate limit history updates and WebKit throws a SecurityError once the cap is hit, which crashed the applications list (argoproj/argo-cd#13156).

goto already builds the final path, so it can compare against history.location directly. Opt-in rather than the default, so existing callers keep their behavior.

This is a follow-up to the review on argoproj/argo-cd#29682 (@choejwoo), which suggested handling this here instead of duplicating the merge logic downstream.

So, this must me merged before we can continue with argoproj/argo-cd#29682


choejwoo (approved):
Thanks for the PR, LGTM!

Since this is opt-in, it should be fine but it could cause issues when removing a fragment like `#example` or opening the same URL in a new tab, right?

rickbrouwer:
Great, thanks @choejwoo.

> Since this is opt-in, it should be fine but it could cause issues when removing a fragment like #example or opening the same URL in a new tab, right?

The new tab one was indeed a bug, the fragment one a behavior difference. Fixed both.

PR #638, opened 2026-09-21
fix: DataLoader stops receiving observable updates after Suspense hide/reveal (#637)

### LINKED ISSUE #637: DataLoader never resubscribes after React 18+ Suspense hide/reveal
`DataLoader` (`src/components/data-loader.tsx`) permanently stops receiving observable emissions after React 18+ hides and re-shows a mounted tree (Suspense effects semantics):`componentWillUnmount` runs on hide (unsubscribes, sets `unmounted = true`), then `componentDidMount` runs again on the same instance with state preserved. But `loadData()` skips resubscribing because `dataWrapper` is already set, and `unmounted` is never reset.

Observed in Argo CD (React 19): collapsing the sidebar writes the preference but the UI no longer reacts, because the view-preferences `DataLoader` lost its subscription.

rickbrouwer:
When a concurrent root (React 18+) hides and re-shows a mounted tree, `componentWillUnmount` tears down the subscription while state is preserved, so `loadData()` never resubscribes and the component stops receiving observable emissions. `componentDidMount` now resets the flag and forces a reload (keeping the current data) when it detects a remount without a subscription.

Includes a regression test that fails without the fix.

Fixes #637


rickbrouwer:
Can you also review this one @choejwoo?
I would like to express my appreciation in advance.

src/components/data-loader.tsx
```diff
@@ -57,7 +57,9 @@ export class DataLoader<D = any, I = undefined> extends React.Component<LoaderPr
     }
 
     public componentDidMount() {
-        this.loadData();
+        this.unmounted = false;
+        const remountedWithoutSubscription = this.state.dataWrapper != null && !this.state.loading && this.subscription == null;
+        this.loadData(remountedWithoutSubscription);
```
choejwoo: It seems like this condition misses observables that have not emitted their first value yet. The loader remains stuck on loading if hiding the tree while `loading === true` and `dataWrapper === null`.

And another case where a successfully resolved Promise also satisfies this condition, so hiding and revealing it calls `load()` again despite the preserved result.

From the code, it looks like these cases may need to be handled.

choejwoo (commented):
Thanks for the PR! I took a look and left a comment. PTAL.

src/components/data-loader.tsx
rickbrouwer: Yeah, true, thanks. Fixed by recording the subscription state in componentWillUnmount instead of inferring it, plus a reload flag for promise results dropped while hidden.

choejwoo (approved):
nit: Completed observables still trigger an extra `load()` on reveal, but it’s not a loop, so it doesn’t seem like a big deal. Could you take another look at that?
Otherwise, LGTM!

And I can’t merge argo-ui PRs, so please ping someone with merge permissions. 😆

rickbrouwer:
Thanks. It should be resolved now.
It’s not just about the merge for me, but also about the good feedback you provide. So, thanks for that.
Hopefully, it will get merged at some point later on 🙂

